PR #1, opened 2016-12-29
adds the identity gate

silky:
it's useful to have the identity gate here.

there is more to be done in terms of making this work with the "IBM Quantum Experience" thing. i'll add that in in a bit, if i figure out how.

### DIFF
--- a/projectq/backends/_ibm/_ibm.py
+++ b/projectq/backends/_ibm/_ibm.py
@@ -24,6 +24,7 @@ from projectq.ops import (NOT,
                           S,
                           Sdag,
                           H,
+                          I,
                           Measure,
                           Allocate,
                           Deallocate,
@@ -87,7 +88,7 @@ class IBMBackend(BasicEngine):
 			return True
 		if get_control_count(cmd) == 0:
 			if (g == T or g == Tdag or g == S or g == Sdag or g == H or g == Y
-			   or g == Z):
+			   or g == Z or g == I):
 				return True
 		if g == Measure or g == Allocate or g == Deallocate:
 			return True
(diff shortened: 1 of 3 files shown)


silky:
it seems as if there is an identity operation that it supports, but i can't find *any* documentation on their API.

maybe the maintainers of this repo have documentation for it? 

in any case, i think adding support for the op within that framework will be very easy when we have the documentation ....

thomashaener:
What would you use it for? I can't think of anything but experimenting with decoherence effects, in which case a Sleep(time)-gate would make more sense in my opinion (the identity operation itself has no duration).

silky:
it's a fairly common idiom to have an identity operation.

i needed this because i was writing something that decides the next operation to execute. it would either set the next operation to be some gate, like `X`, or the identity operation, `I`.

it makes code cleaner

```
if condA:
  op = I

if condB:
  op = X

if condC:
  op = Z

op | qubit
```

if just much cleaner than:

```
if not(condaA):
  if condB:
    op = X
  if condC:
    op = Z

  op | qubit
```

thomashaener:
Is only one of your condition variables True? If not then the following does the same as your first code (minor modifications are necessary to get the behavior of your 2nd code) and I find it at least as clean:
```
if condC:
  Z | qubit
elif condB:
  X | qubit
```
Also, if your condition variable depends on a measurement outcome, you can use
```
with Control(eng, measured_qubit):
  ...
```
which will allow to draw the circuit with classical control lines and to postpone the measurement until after the controlled operation has been executed if the hardware back-end does not support fast feedback (effectively performing, e.g., a 'quantum-controlled' X instead of classical branching).

PR #15, opened 2017-03-06
Simplify angle comparison in BasicRotationGate.__eq__

Strilanc:
Use a shifted mod to do the comparison. Also use <= instead of <, since < makes no sense for tolerance=0.


Strilanc:
Is there a way to configure this coverage tool to give incremental coverage results instead of global ones?

Just deleting covered lines will decrease coverage. Because of that, I don't care about the global coverage percentage going down. What I care about is not introducing uncovered lines.

Strilanc:
Oh right. I've gotten too used to $U^{t}$ instead of $e^{i U t}$...

Strilanc:
Fixed the periodicity and added a test checking it.

projectq/ops/_basics.py
```diff
@@ -295,23 +295,29 @@ def get_merged(self, other):
...
-                return True
-            else:  # if one angle is close to 0 and the other close to 4*pi:
-                if difference > 4 * math.pi - tolerance:
-                    return True
-                else:
-                    return False
-        else:
+        if not isinstance(other, self.__class__):
             return False
+        difference = abs(_signed_mod(self._angle - other._angle, 4 * math.pi))
```
damiansteiger: What is the advantage of this construction?
I don't think it simplifies this function. Calculating a signed mod but then taking the abs()...

projectq/ops/_basics.py
```diff
@@ -295,23 +295,29 @@ def get_merged(self, other):
...
-        if isinstance(other, self.__class__):
-            difference = abs(self._angle - other._angle)
-            if difference < tolerance:
-                return True
-            else:  # if one angle is close to 0 and the other close to 4*pi:
-                if difference > 4 * math.pi - tolerance:
-                    return True
-                else:
-                    return False
-        else:
```
damiansteiger: How about we just remove the last to `else:` and just add here `return False` to simplify the function?

projectq/ops/_basics.py
```diff
@@ -295,23 +295,29 @@ def get_merged(self, other):
 
     def __eq__(self, other):
         """ Return True if same class and same rotation angle. """
-        tolerance = 1.e-12
-        if isinstance(other, self.__class__):
-            difference = abs(self._angle - other._angle)
```
damiansteiger: The precondition of this class is that the angle is in the interval [0, 4*pi). Your construction made it "safer" as it still worked if that was not the case. We can easily add this here by
`difference = abs(self._angle - other._angle) % (4 * math.pi)`

projectq/ops/_basics_test.py
```diff
@@ -181,6 +181,7 @@ def test_basic_rotation_gate_comparison():
     # Test different types of gates
     basic_gate = _basics.BasicGate()
     assert not basic_gate == basic_rotation_gate6
+    assert basic_rotation_gate2 != _basics.BasicRotationGate(0.5 + 2 * math.pi)
```
damiansteiger: Thanks for this test. We forgot the add this this one

damiansteiger:
Let me know what you think about my comments and suggested implementation. If you agree to them, let me know if you want me to make these changes or if you will make them.
Thanks

projectq/ops/_basics.py
Strilanc: That will give the wrong final result when self._angle is 4pi-epsilon and other._angle is zero, The difference will end up being large instead of small, despite the angles being close.

damiansteiger: It won't give the wrong result. There are two cases when this `__eq__` should return true. Case 1: 
`if difference < tolerance:`
and case 2:
`if difference > 4 * math.pi - tolerance:`
both of these cases are handled correctly

Strilanc: Oh I see, you'd keep the double comparison.

The advantage of signed_mod is that it puts close differences near zero, so you don't have to remember to do two comparisons. It wraps you into [-c/2, c/2] instead of [0, c]. This is almost always what you want when doing comparisons, or when deciding "which direction should I turn to get there?" when working modulo a constant.

damiansteiger: Yes, I would keep the double comparison instead of introducing another function call. But I would take your suggestions that it would be better `difference = ... % (4 * math.pi)` as with the modulo operation we don't rely on the angle being within [0,4*math.pi) but that it works generally. Also a small simplification of the original code can be done by removing the last two
```
else: 
   return False
else:
   return False
```
to just
`return False` if none of the above two `if` were true.

Issue #21, opened 2017-03-13
Decomposition rules being global will make testing and configuration harder


Strilanc:

Strilanc:
Fixed by #22

PR #22, opened 2017-03-14
Deglobalizing decomposition rule sets

Strilanc:
With decomposition rules being added to a global config when imported, it is extremely hard to tell how they are getting into a given test. This pull adds a DecompositionRuleSet member to AutoReplacer, uses that instead of the global, and deals with all the consequences of that.

Clearly this is only an interim solution. Once there are dozens or hundreds of decompositions it will get unwieldly to manually add them all. There will need to be *some* kind of packaging system for the rules, but the import approach is very difficult to control and test because of its idempotence and global effects.


Strilanc:
(Also, this is another "trying things out" PR that's not ready for serious review yet.)

thomashaener:
I completely agree with testing being harder if decompositions are global. Having a non-global alternative which still doesn't require the user to worry too much about these dicts would be great.

Strilanc:
I started working on it in https://github.com/ProjectQ-Framework/ProjectQ/pull/22

Strilanc:
For https://github.com/ProjectQ-Framework/ProjectQ/issues/21

Strilanc:
Should come after https://github.com/ProjectQ-Framework/ProjectQ/pull/24#issuecomment-286917849

Strilanc:
This is ready for review now.

There's space indentation trying to sneak in via pycharm's autoformatting and causing diff-noise, but that will be resolved once #24 is merged (or fixed after #24 is rejected).

thomashaener:
Thanks a lot! Could you add the decompositions to the shor.py example as well? (we should also add that example to the hook).
Also, I think it would be nicer for users if we still had something similar to the import statement from before, even if this requires global variables. Instead of using `import`, we could have something like
```
projectq.setups.use("ibm")
```
which sets the default engines & decompositions (for the replacer). When testing, we can call
```
projectq.setups.use("none")
```
Internally, this uses your solution.

Strilanc:
@thomashaener I'm very wary of any statement that has global effects like that. It breaks encapsulation between modules.

A step up would be something like this:

```python
with projectq.setups.use("ibm"):
    stuff
```

because at least this scopes the global effects, so they can't be accidentally triggered when testing. But even that can amplify the effects of bugs that let the state escape.

I think we should be using a proper dependency injection framework for this kind of configuration. Step one is to actually pass the configuration around, then that gets tiring, then we inject.

thomashaener:
@Strilanc Oh my bad, nevermind the ' & decompositions', I don't care about those. But the current solution still uses the import statement to set the default engines. I thought we wanted to change that as well, since it makes testing the setups harder. So instead of
```
import projectq.setups.default
```
I was suggesting to have a function to set `projectq.default_engines` (and those are already equipped with default decomposition rules, so that's fine). If users go through the trouble of defining their own compiler engine list, then they can also deal with passing the decomposition rules to the replacer, I am not worried about that.

The nicer solution would be to have them import the default engines from the setup they want to use and then always write
```
from projectq.setups.default import default_engines
...
eng = MainEngine(engine_list=default_engines())
```
but I don't think many users would agree that this is nicer :-)

Strilanc:
Oh I see. Yes, it's also worth moving that config out of the imports as well. But this PR will focus just on the ruleset.

Strilanc:
This is ready for review now.

examples/shor.py
```diff
@@ -104,8 +108,10 @@ def high_level_gates(eng, cmd):
 if __name__ == "__main__":
     # build compilation engine list
     resource_counter = ResourceCounter()
-    compilerengines = [AutoReplacer(), InstructionFilter(high_level_gates),
-                       TagRemover(), LocalOptimizer(3), AutoReplacer(),
+    dh = DecompositionRuleSet(modules=[projectq.libs.math,
```
thomashaener: We should probably use a better name than `dh`, at least in the example codes.

projectq/cengines/_replacer/_decompositionhandling.py
```diff
@@ -11,67 +11,101 @@
...
-decompositions = dict()
 
-
-class ThisIsNotAGateClassError(Exception):
+class ThisIsNotAGateClassError(TypeError):
     pass
 
 
-def register_decomposition(gate_class, gate_decomposer, gate_recognizer=None):
+class DecompositionRule:
```
thomashaener: `DecompositionRule` seems very similar to `Decomposition`, so I'm wondering if we want to either merge the two or make `all_defined_decomposition_rules`, e.g., a dict of lists of Decompositions, so we can get the gate class when iterating through `all_defined_decomposition_rules`.

examples/shor.py
thomashaener: Also, we'll need decomposition rules in the jupyter notebook (in examples/) or, alternatively
```
compilerengines = projectq.default_engines() + [CommandPrinter()]
```
(plus an `import projectq` somewhere).

Strilanc: I renamed the `dh`'s to `rule_set`, and also renamed a couple files.

I checked that all the examples run (except the ones that require connecting to IBM).

projectq/cengines/_replacer/_decompositionhandling.py
Strilanc: Changed `Decomposition` to a private class inside `_decomposition_rule_set.py`.

thomashaener: @Strilanc I'm sorry about the long wait, github hid this thread from me.

I still think it would be nicer if `DecompositionRuleSet` actually contained `DecompositionRule` objects (instead of `_Decomposition` objects). Why not combine them into one class?

Strilanc: Well, DecompositionRule requires a gate class for indexing purposes but there's currently no way to compute the inverse class when construction the inverse DecompositionRule.

I'm not too worried about it. `_Decomposition` is a private class, and I generally go by the rule that anything you do in the privacy of your own module is fine by me. It's easy to change stuff that isn't coupled.

thomashaener: Fair enough.

Issue #12, opened 2017-03-06
Tabs being mixed with spaces


Strilanc:
Many files in the project have spaces hidden amongst the tabs. Sometimes by accident, sometimes on purpose for fine-grained alignment (despite that not making any sense with tabs present).

Given that PEP8 strongly recommends spaces, we should probably switch to spaces.

damiansteiger:
We started out by using tabs because it give the user more freedom. You can either set the tab width to be 2 spaces or 4 spaces in your editor of choice.

If you then want fine-grained alignment, it needs to be mixed: As many tabs as the line above and then spaces s.t. the alignment is correct for both tab width 4 or tab width 2.

We are currently considering to move to spaces, so thanks for your input.

In any case, we should add an automatic linter to check all new commits and clean up the old code. I am not sure if there is a linter which can handle tabs + spaces (for fine-grained alignment)...

maffoo:
+1 for spaces in python code. They're essentially universal for python, and using tabs will likely create friction for python devs. For c++ code tabs are fine, though my personal preference would be for spaces there too... :-)

damiansteiger:
@Strilanc @maffoo: As we are deciding this issue in the next days, we would like to get your input:

I understand that you both have a strong preference for spaces over tabs. Any preference on 4 vs 2 spaces? (I currently show my tabs as 4 spaces)

Also while we are already on it, we can also discuss line width. While PEP says 79 characters, I often find this a bit short...

Strilanc:
4 spaces is more common, and your existing alignments are set up based on tabs being 4 spaces (making the transition trivial), so I would go with 4.

I don't have any strong line width preferences. My preference literally changes monitor by monitor. I guess I wouldn't go lower than 79 or higher than 160. For a physics thing, 137 is kind of tempting...

thomashaener:
I'm currently using 2 spaces, but our alignments work with any tab-width; this is why we mixed tabs and spaces for aligning multi-line statements:
```
...
\t\tsome_function_with_many_parameters(param1,
\t\t                                   param2,
...
```
I would be fine using 4 spaces, though.

As for the width: 137 would indeed be tempting :) but it's probably a bit too long; I'm ok with PEP8.

Strilanc:
Oh, you were careful that all semantic indentation was tabs and all alignment was spaces? Then nevermind the point about preferring 4 for that reason.

maffoo:
[...]

PR #24, opened 2017-03-15
Convert .py indentation from tabs to spaces and remove trailing spaces

Strilanc:
To produce the changes in this PR I ran exactly these three bash commands:

    files=$(find | grep \.py$)

    for file in $files; do sed -i 's/\t/    /g' ${file}; done

    for file in $files; do sed -i 's/ *$//g' ${file}; done

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -294,7 +294,7 @@ latex_documents = [
 #
 # latex_appendices = []
 
-# It false, will not define \strong, \code, 	itleref, \crossref ... but only
+# It false, will not define \strong, \code,     itleref, \crossref ... but only
 # \sphinxstrong, ..., \sphinxtitleref, ... To help avoid clash with user added
 # packages.
 #
--- a/examples/grover.py
+++ b/examples/grover.py
@@ -6,75 +6,75 @@ from projectq.meta import Loop, Compute, Uncompute, Control
 
 
 def run_grover(eng, n, oracle):
-	"""
-	Runs Grover's algorithm on n qubit using the provided quantum oracle.
-	
-	Args:
-		eng (MainEngine): Main compiler engine to run Grover on.
-		n (int): Number of bits in the solution.
-		oracle (function): Function accepting the engine, an n-qubit register, and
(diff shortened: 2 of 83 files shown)


Strilanc:
For https://github.com/ProjectQ-Framework/ProjectQ/issues/12

damiansteiger:
Thanks a lot. We have not decided about spaces vs tabs yet. Anyway we will change it once we integrate FermiLib into ProjectQ.

Strilanc:
@damiansteiger Integrating FermiLib is going to take months. There's no reason to block a tiny change like this on a large library like that.

The tab and space mixing causes me minor problems every time I touch the code. Pycharm can't figure out if I want lines indented with tabs or spaces, so it just kinda chooses arbitrarily. Differing tab widths mean existing code doesn't line up argument-by-argument like it was intended to. I can never tell if something will line up on github or not before actually doing a push.

Just make a decision about whether to do the spaces conversion or not. I'm in favor of spaces for the same reasons outlined in PEP8. But mostly I'm in favor of not unnecessarily pushing off the decision.

Strilanc:
Note that by "in favor" of spaces I mean *strongly* in favor. PEP8 should win unless there's a good reason not to use it.

And if we're using tabs, we need to actually ban fine alignments and set up automatic lint checks to catch spaces sneaking in. A lot of tools insert spaces when you split code between lines.

damiansteiger:
Fair enough, we increased priority of this issue and will decide and fix it in the next days.

By the way: Integrating FermiLib will not take months ;)

damiansteiger:
Thanks for your suggestion to move to spaces instead of tabs. We have decided to change indentation to 4 spaces as suggested and hope that it will make it easier for developers to contribute. While our previous indentation (tabs for semantic indentation and spaces for the rest) worked well, it did create more confusion than necessary, see e.g. #12.

@thomashaener took time to not just change the indentation but at the same time also enforce a the line width to 79 character in order to be fully PEP8 compatible, see pull request #29. Therefore, I am closing this pull request.

Issue #37, opened 2017-04-11
Controlled tensors work inconsistently w.r.t. non-controlled tensors


Strilanc:
When you apply:

    All(X) | targets

The engine receives X commands via the __or__ implementation in Tensor.

But if you apply:

    C(All(X)) | (control, targets)

Then then the __or__ operation is bypassed and the engine receives a Tensor command instead.

Strilanc:
I believe the fix here is to replace `generate_command` with `generate_commands`, fix up any existing usage to deal with the multiplicity, and always always have `__or__` delegate to `generate_commands`.

damiansteiger:
Good catch. I will have a look at it

damiansteiger:
This is addressed in #53

PR #40, opened 2017-04-11
[Reworking] Fix inconsistency between __or__ and generate_command

Strilanc:
- Stacking modifiers was bypassing lower modifiers in some cases
- Added controlled-tensor test to check that one of the cases is fixed
- Added 'restart_recording' to DummyBackend
- Fixed Qubit.__eq__ not being symmetric
- Refactored generate_command into generate_commands
- Replaced custom __or__'s with custom generate_commands
- Added type annotations in several places
- Added __repr__ methods to several classes
- Added 'controls' parameter to Command.__init__, and used it in several places

### DIFF
--- a/projectq/ops/_metagates.py
+++ b/projectq/ops/_metagates.py
@@ -166,6 +166,10 @@ class ControlledGate(BasicGate):
         """ Return string representation, i.e., CC...C(gate). """
         return "C" * self._n + str(self._gate)
 
+    def __repr__(self):
+        return ("ControlledGate(" + repr(self._gate) + ", " +
+                repr(self._n) + ")")
+
     def get_inverse(self):
         """
         Return inverse of a controlled gate, which is the controlled inverse
@@ -173,7 +177,7 @@ class ControlledGate(BasicGate):
         """
         return ControlledGate(get_inverse(self._gate), self._n)
 
-    def __or__(self, qubits):
+    def generate_commands(self, qubits):
         """
         Apply the controlled gate to qubits, using the first n qubits as
         controls.
@@ -203,9 +207,11 @@ class ControlledGate(BasicGate):
             raise ControlQubitError("Wrong number of control qubits. "
                                     "First qureg(s) need to contain exactly "
                                     "the required number of control qubits.")
-        cmd = BasicGate.generate_command(self._gate, tuple(gate_quregs))
-        cmd.add_control_qubits(ctrl)
-        apply_command(cmd)
+        result = []
+        for cmd in self._gate.generate_commands(tuple(gate_quregs)):
+            cmd.add_control_qubits(ctrl)
+            result.append(cmd)
+        return result
 
     def __eq__(self, other):
         """ Compare two ControlledGate objects (return True if equal). """
@@ -254,6 +260,9 @@ class Tensor(BasicGate):
         """ Return string representation. """
         return "Tensor(" + str(self._gate) + ")"
 
+    def __repr__(self):
+        return "Tensor(" + repr(self._gate) + ")"
+
     def get_inverse(self):
         """
         Return the inverse of this tensored gate (which is the tensored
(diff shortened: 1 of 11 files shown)


projectq/types/_qubit.py
```diff
@@ -83,7 +86,8 @@ def __eq__(self, other):
         Args:
             other (BasicQubit): BasicQubit to which to compare this one
         """
-        return (isinstance(other, self.__class__) and self.id == other.id and
```
Strilanc: Using `isinstance(other, self.__class__)` means that `a == b` can return a different result than `b == a` when `b` is a subclass of `a`. Equality must *must* be commutative, so I changed `self.__class__` to `BasicQubit`.

thomashaener:
What do you mean by "Stacking modifiers was bypassing lower modifiers in some cases"?

damiansteiger:
Is this also work in progress as the tests are failing?

Strilanc:
@thomashaener https://github.com/ProjectQ-Framework/ProjectQ/issues/37#issuecomment-293492619

It looks like the test failure is because a test was too specific. I'll look into it.

Strilanc:
Ah, I just needed to make the controls weakrefs.

projectq/ops/_metagates.py
```diff
@@ -166,14 +166,18 @@ def __str__(self):
         """ Return string representation, i.e., CC...C(gate). """
         return "C" * self._n + str(self._gate)
 
+    def __repr__(self):
+        return ("ControlledGate(" + repr(self._gate) + ", " +
+                repr(self._n) + ")")
```
maffoo: Consider using `format`:
```python
return "ControlledGate({!r}, {!r})".format(self._gate, self._n)
```

Strilanc:
@damiansteiger You mentioned we should override __or__ instead of generate_commands? Could you go into the details of why that is needed in FermiLib?

damiansteiger:
We want to be able to adjust the syntax for specific gates. 
In case of FermiLib, it would be nice to create a Hamiltonian, e.g., X0X5, and use a syntax such as
```
wavefunction = eng.allocate_qureg(6)
TimeEvolution(t, hamiltonian) | wavefunction # hamiltonian being an object representing X0X5
```
This will create a gate acting on 6 qubits, until a first decomposition rule will figure out that this is actually only a gate acting on 2 qubits.
One way to solve this would be to write
```
TimeEvolution(t, h_short) | (wavefunction[0], wavefunction[5]) # h_short representing X0X1
```
This will only create a two qubit gate. So we should allows the user to use the first syntax as it is much nicer and internally we create h_short and apply it only to two qubits.

Previous examples of additional syntax are `All()` and `C()`. I can think of two ways to get there:

1. We keep the `|` operator identical for all gates but add an additional function `apply_gate(...)` which is called by `|` and which every gate can change to create the desired syntax. Internally it calls `generate_command(...)`. I wouldn't put everything into a new `generate_commands` as one function. `apply_gate` should handle the syntax and `generate_command` should only generate commands (and not be adapted for different gates). Note that this would add an additional function call.

2. We keep `generate_command` identical for all gates but adjust the `|` operator to allow for the syntax which we want to have for, e.g., `TimeEvolution` or `C()`. To implement `C` or `All` and allowing them to be chained, e.g., `C(All(X))`, there is a shallow implementation solution which requires almost no changes #53 by making `C(X)` create a ControlledGate whose `|` operator calls the `|` operator of the gate, i.e. `|` of `X`. No additional function call for normal gates. Also in this version, the `|` operator handles all the syntax.

Any opinions on these two different approaches?

I would prefer the second method for performance reasons.

damiansteiger:
Btw: Next time, I would suggest to split up such pull requests so that we can work on independent things in parallel, e.g., the adding `__repr__` method at several places would be a fast approval.

Strilanc:
I don't understand why having `__or__` feed the result of a `generate_commands` method (note the S at the end) into an engine prevents the use case you described. Just override the `generate_commands` on `TimeEvolution` to do logic that returns a 2-qubit command instead of a 6-qubit command.

thomashaener:
@damiansteiger @Strilanc I guess there was a misunderstanding concerning what `generate_command` is supposed to do. While we could introduce another function which gets called by `__or__`, this would introduce an extra function call and I don't see any issue with overriding `__or__` directly for specialization which outweighs the performance issue.

Issue #43, opened 2017-04-12
BasicQubit hash is inconsistent


Strilanc:
The most important condition that hash functions must satisfy is `a == b implies hash(a) == hash(b)`, but the BasicQubit class doesn't meet it:

    a = _qubit.BasicQubit(fake_engine, 5)
    b = _qubit.BasicQubit(fake_engine, 5)
    assert a == b
    assert hash(a) == hash(b)

The existing tests are checking for inequality instead of equality, i.e. that `a != b implies hash(a) != hash(b)`. Which is actually not a requirement of hash functions at all, though it matters for performance.

thomashaener:
While the hash would remain inconsistent, we could add support for all cases where `id != -1` without any further changes.
Alternatively, we can substitute the `MainEngine`'s WeakSet with some other data structure...

PR #44, opened 2017-04-12
Fix BasicQubit.__hash__ not satisfying a==b implies hash(a)==hash(b)

Strilanc:
Fixes https://github.com/ProjectQ-Framework/ProjectQ/issues/43

### DIFF
--- a/projectq/types/_qubit_test.py
+++ b/projectq/types/_qubit_test.py
@@ -75,6 +75,7 @@ def test_basic_qubit_hash():
     assert len(set(hash(_qubit.BasicQubit(fake_engine, e))
                    for e in range(100))) == 100
 
+    # Important that weakref.WeakSet in projectq.cengines._main.py works.
     # When id is -1, expect reference equality.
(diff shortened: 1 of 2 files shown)


projectq/types/_qubit_test.py
```diff
@@ -60,21 +60,26 @@ def test_basic_qubit_comparison(id0, id1, expected):
...
 
 
 def test_basic_qubit_hash():
     fake_engine = "Fake"
-    qubit0 = _qubit.BasicQubit(fake_engine, 0)
-    qubit1 = _qubit.BasicQubit(fake_engine, 1)
-    assert hash(qubit0) != hash(qubit1)
-    qubit0.id = -1
-    qubit1.id = -1
-    # Important that weakref.WeakSet in projectq.cengines._main.py works.
```
damiansteiger: This is an important comment and still valid afterwards, hence deleting it is not allowed

Strilanc: Done.

PR #49, opened 2017-04-14
Fix ControlledGate failing when a register is empty

Strilanc:
- Fixed make_tuple_of_qureg assuming first register will contain items
- Fixed ControlledGate overriding __or__ instead of generate_command
- Added test of n=0 case

### DIFF
--- a/projectq/ops/_basics.py
+++ b/projectq/ops/_basics.py
@@ -169,12 +169,10 @@ class BasicGate(object):
         """
         qubits = self.make_tuple_of_qureg(qubits)
 
-        for i in range(len(qubits)):
-            for j in range(len(qubits[i]) - 1):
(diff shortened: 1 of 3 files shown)


Strilanc:
This is ready for review. @damiansteiger

Strilanc:
Now blocked by issue https://github.com/ProjectQ-Framework/ProjectQ/issues/56

Strilanc:
Ready for review.

thomashaener:
Thanks! Now n=0 fails if there is no control register, but I think this should be fine since there's a nice exception explaining what went wrong... or do we want to support both cases?

Strilanc:
I think the correct behavior, when all registers are empty, is usually to do nothing. But that is more likely to have some exception I haven't thought of, and can be done in a separate PR.

PR #51, opened 2017-04-15
Add ClassicalSimulator for faster testing of non-quantum circuits

Strilanc:


Strilanc:
Comes after https://github.com/ProjectQ-Framework/ProjectQ/pull/49

projectq/backends/_sim/_classical_simulator.py
```diff
@@ -0,0 +1,131 @@
...
+        Reads a group of bits as a little-endian integer.
+
+        Args:
+            qureg (projectq.types.Qureg):
+                The group of bits to read, in little-endian order.
+
+        Returns:
+            int: Little-endian register value.
+        """
+        return sum(self.read_bit(qureg[i]) << i
```
thomashaener: I'd suggest using `result |= (value << i)` in a loop (it's almost 2x faster than summation)

projectq/backends/_sim/_classical_simulator.py
```diff
@@ -0,0 +1,131 @@
...
+
+        if isinstance(cmd.gate, XGate):
+            assert len(cmd.qubits) == 1 and len(cmd.qubits[0]) == 1
+            target = cmd.qubits[0][0]
+            if all(self.read_bit(c) for c in cmd.control_qubits):
+                self.write_bit(target, not self.read_bit(target))
+            return
+
+        if isinstance(cmd.gate, BasicMathGate):
+            if all(self.read_bit(c) for c in cmd.control_qubits):
```
thomashaener: For the controls I'd suggest building a mask and then compare `mask & state == mask`.

thomashaener:
I added a few suggestions to improve performance. I think these will help when testing large circuits (for, e.g., Shor's algorithm). I will also add a C++ version as soon as I find time.

projectq/backends/_sim/_classical_simulator.py
```diff
@@ -0,0 +1,131 @@
+"""
```
damiansteiger: Missing license header:

Every new file should start with the same Apache license boiler plate.

Strilanc: Done.

projectq/backends/_sim/_classical_simulator.py
Strilanc: Done.

projectq/backends/_sim/_classical_simulator.py
Strilanc: Done.

Strilanc:
Note: still comes after #49

thomashaener:
Oh I think the issue was the PR title which is different from what the PR actually addresses: It fixes the case where there is an empty control register, rather than n=0. Controlling with n=0 works fine as long as there is no empty register.

projectq/backends/_sim/_classical_simulator.py
```diff
@@ -0,0 +1,165 @@
...
+            assert len(cmd.qubits) == 1 and len(cmd.qubits[0]) == 1
+            target = cmd.qubits[0][0]
+            if meets_controls:
+                self.write_bit(target, not self.read_bit(target))
+            return
+
+        if isinstance(cmd.gate, BasicMathGate):
+            if meets_controls:
+                ins = [self.read_register(reg) for reg in cmd.qubits]
+                outs = cmd.gate.get_math_function(None)(ins)
```
thomashaener: get_math_function needs cmd.qubits (if users override its behavior such as in the AdvancedBasicMath gate you proposed).

Strilanc: Done.

thomashaener:
Great, thanks!

damiansteiger:
👍

PR #57, opened 2017-04-18
Add 'controls' and 'tags' parameters to Command.__init__

Strilanc:
And use them in code that was following up the constructor with modifications to those fields.


Strilanc:
@thomashaener This is ready for review.

PR #58, opened 2017-04-19
[Later] ImprovedSuggestion: Implement _GateWithIncludedControls accessed with `gate & control` syntax

Strilanc:
Adds ampersand-as-control syntax:

    gate & controls | targets
    gate & one_control & more_controls | targets

The advantages of this approach over `ControlledGate` and `with Control` are:

- Don't have to redundantly enter the number of controls.
- Succinct. Doesn't require imports or `with`.
- Easier to check at a glance. Controls are inline but visually separate from the other qubits.

This PR comes after PR https://github.com/ProjectQ-Framework/ProjectQ/pull/57


damiansteiger:
Thanks for your suggestion. I don't think the syntax is too complicated writing
```
C(gate, len(controls)) | (controls, targets)
# or
with Control(eng, controls):
    gate | targets
```
We see the following conflicts regarding your proposal with our current syntax design:
1.  The syntax of ProjectQ is supposed to be close to physics/math notation, which would be 
```
C(gate) |controls> |targets> = C(gate) | controls, targets >
```
and therefore the above syntax `C(gate) | (controls, targets)`

2. One fundamental rule we don't want to abandon (consistency) is that everything left of `|` are classical arguments.

3. `C()` is merely used to create shortcuts, such as `CNOT=C(NOT)` and `Toffoli=C(NOT,2)`. User code should prefer `with Control()`. It is hence not an integral part of the syntax and therefore we don't see the advantage of abandoning the main syntax design.

PR #69, opened 2017-04-23
Feature/QubitOperator

damiansteiger:
Adding a class to store an operator (sum of tensor products of Pauli operators) acting on qubits.


damiansteiger:
Please do **not** squash  and merge this pull request.

PR #71, opened 2017-04-26
Support Qureg.__int__ for multiple bits (little-endian)

Strilanc:
I think this is the natural thing to do when asking for the `int` in a quregister. (Existing tests weren't changed because they happened to not check for this.)

A reasonable follow-up change would be to remove the len==1 constraint from `__bool__` and `__non_zero__`, and have them check if `__int__` is non-zero instead. I can do that in this PR if you want.

### DIFF
--- a/projectq/types/_qubit.py
+++ b/projectq/types/_qubit.py
@@ -187,11 +187,7 @@ class Qureg(list):
             Exception if more than 1 qubit resides in this register (then you
             need to specify which value to get using qureg[???])
         """
-        if len(self) == 1:
-            return int(self[0])
-        else:
-            raise Exception("__int__(qureg): Quantum register contains more "
-                            "than 1 qubit. Use __bool__(qureg[idx]) instead.")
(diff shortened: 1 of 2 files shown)


thomashaener:
This is actually what our first implementation did, but we decided to drop this feature.

Rather than adding this functionality to `Qureg`, I would suggest to add a new quantum type `Quint` (and eventually also `Qufixed, Qufloat, ...`). This has the advantage that such a conversion must be performed explicitly (it's a different interpretation of a `Qureg`after all): `a = Quint(qureg1)` and it would also behave as expected in terms of operations:
```
...
a = Quint(qureg1)
b = Quint(qureg2)
a += b # as opposed to concatenation: qureg1+qureg2
Measure | a
m_a = int(a) # is actually a+b
```
I think this explicit conversion helps avoid bugs such as `if bool(qureg): ...`, where users might have wanted `all(bool(q) for q in qureg)`.

PR #72, opened 2017-04-27
Add a first version of time evolution emulation

thomashaener:
This is the first version of the time evolution emulation feature of the simulator. It works but uses a very loose estimate to rescale the operator H when calculating exp(H). This results in long runtimes for operators featuring a large 1-norm. Later: Include tighter bounds by Al-Mohy & Higham, 2011.

### DIFF
--- a/projectq/backends/_sim/_pysim.py
+++ b/projectq/backends/_sim/_pysim.py
@@ -251,7 +251,8 @@ class Simulator(object):
     def emulate_time_evolution(self, terms_dict, time, ids, ctrlids):
         """
         Applies exp(-i*time*H) to the wave function, i.e., evolves under
-        the Hamiltonian H for a given time.
+        the Hamiltonian H for a given time. The terms in the Hamiltonian
+        are not required to commute.
 
         This function computes the action of the matrix exponential using
         ideas from Al-Mohy and Higham, 2011.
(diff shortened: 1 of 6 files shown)


projectq/backends/_sim/_pysim.py
```diff
@@ -239,22 +239,60 @@ def get_expectation_value(self, terms_dict, ids):
...
+            self._apply_term(term, ids)
             delta = coefficient * _np.vdot(current_state, self._state).real
             expectation += delta
             self._state = _np.copy(current_state)
         return expectation
 
+    def emulate_time_evolution(self, terms_dict, time, ids, ctrlids):
+        """
+        Applies exp(-i*time*H) to the wave function, i.e., evolves under
+        the Hamiltonian H for a given time.
```
damiansteiger: Maybe add that this function is general and allows the Hamiltonian to be a sum of non commuting terms

projectq/backends/_sim/_simulator.py
```diff
@@ -97,7 +98,8 @@ def is_available(self, cmd):
         """
         if (cmd.gate == Measure or cmd.gate == Allocate
            or cmd.gate == Deallocate
-           or isinstance(cmd.gate, BasicMathGate)):
+           or isinstance(cmd.gate, BasicMathGate)
+           or isinstance(cmd.gate, TimeEvolution)):
```
damiansteiger: Do we always want to use the general emulation feature for Hamiltonians with non-commuting terms even if the Hamiltonian has only one term?

damiansteiger: Never mind. Users can use an instruction filter engine to decompose Hamiltonians with only a single term if the decomposition to one and two-qubit gates is faster to simulate.

PR #89, opened 2017-05-06
[Blocked on #93] Remove debug 'on error print and continue' code in LocalOptimizer._send_qubit_pipeline

Strilanc:
Continuing the execution is useful when debugging, but disastrous when trying to achieve some task. The corrupted state will spread to other parts of the program, and potentially into output data.

### DIFF
--- a/projectq/cengines/_optimize.py
+++ b/projectq/cengines/_optimize.py
@@ -56,22 +56,18 @@ class LocalOptimizer(BasicEngine):
                 for qb in qreg:  # ... and qubits involved
                     Id = qb.id
                     if Id != idx:  # is a different qubit than the current one
-                        try:
-                            gateloc = 0
-                            # find location of this gate within its list
-                            while self._l[Id][gateloc] != il[i]:
-                                gateloc += 1
+                        gateloc = 0
+                        # find location of this gate within its list
+                        while self._l[Id][gateloc] != il[i]:
+                            gateloc += 1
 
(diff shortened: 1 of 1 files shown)


thomashaener:
The problem is that removing this will cause huge error messages when the program is terminated early (e.g., ctrl+C).

Strilanc:
If the program is throwing errors on shutdown, that indicates you have a bug in a __del__ or atexit method somewhere.

Do you have a test that triggers the buggy __del__ code path, so I can fix that first?

thomashaener:
No, this is not at all the issue. If you hit ctrl-C while the optimizer is saving gates, then that operation is stopped (and certain multi-qubit gates may not have been added to all qubits they act upon).
The garbage collection will now cause `Deallocate` gates being sent down the pipeline, which will in turn flush the buffer of the optimizer. During this process, this exception can occur.

Strilanc:
That's a classic re-entrancy bug. If `receive` methods are not safe to re-enter while they are executing, calling them in a `__del__` method is a bug. The fix is either to ensure `receive` is not called in `__del__`, or to rewrite `receive` to be re-entrant safe (e.g. by using a queuing mechanism).

thomashaener:
So you'd want to remove automatic deallocation of qubits?

Strilanc:
I think you guys want that quite a lot, so I'll look into adding a queuing mechanism instead.

thomashaener:
Indeed :)

PR #90, opened 2017-05-06
Minor cleanup in LocalOptimizer._optimize

Strilanc:
- Use slicing and a comprehension instead of index twiddling to count identical gates
- Separate match-finding from match-picking


thomashaener:
Coveralls is currently down. We'll see if they fix it; if not, we'll use another tool.

PR #92, opened 2017-05-06
Simplify BasicEngine.deallocate_qubit

Strilanc:
- Use guard statement instead of indenting
- Specify dirty tag via `Command.__init__` instead of via CommandModifier
- Merge clean/dirty cases.


Strilanc:
Something about putting the Qubit into a Qureg seems to be causing.. key errors? Not in the tests, but when injecting failures into some LocalOptimizer methods.

Issue #93, opened 2017-05-07
[WIP] Fix Qubit.__del__ causing problems during abnormal shutdown


Strilanc:
Based on to the answer to https://github.com/ProjectQ-Framework/ProjectQ/issues/85, the reason `MainEngine.__del__` exists and the reason `atexit` is used in MainEngine is to avoid issues caused by qubits being deallocated during sudden shutdowns (e.g. user hitting ctrl+C).

There are actually several places in the code doing acrobatics to work around the issue. And anytime a test fails, you tend to see a bunch of extra failures get tacked on due to the qubit deallocations running over code that failed halfway through in a bad state.

For example, here is a failing repro test:

```
def test_qubit_deallocate_reentrancy_safety():
    class InjectedBugEngine(_basics.BasicEngine):
        def __init__(self):
            _basics.BasicEngine.__init__(self)
            self.was_in_middle_of_receive = False
            self.x = []

        def receive(self, command_list):
            assert not self.was_in_middle_of_receive
            self.was_in_middle_of_receive = True

            for cmd in command_list:
                assert not isinstance(cmd.gate, DeallocateQubitGate)
                if not isinstance(cmd.gate, AllocateQubitGate):
                    self.x[1] = None  # Trigger IndexError

            self.was_in_middle_of_receive = False

    bug_eng = InjectedBugEngine()
    eng = MainEngine(backend=bug_eng, engine_list=[])

    try:
        q = eng.allocate_qubit()[0]
        try:
            H | q  # Triggers injected IndexError bug.
        finally:
            q.__del__()  # Force dealloc of q, as if it went out of scope.
    except IndexError:
        pass  # Expected.

    assert bug_eng.was_in_middle_of_receive
```

This test fails not once but *twice*. Both times on the `assert not self.was_in_middle_of_receive` line. One due to the manual call to `q.__del()__`, and another in `atexit` while the interpreter is shutting down.

To fix this problem, we need to make engines re-entrant safe against qubit deallocations. I suggest introducing a `defer_deallocate_qubit` method that puts the deallocate command into a queue. Then, before each non-reentrant call to receive, the engine drains that queue.

So, during shutdown, the command will go into a queue and just not do anything. But during actual use, the user will call Allocate or Flush or `H | something`, and the deallocate will appear in the same order as usual. The main change will be that circuits may be missing deallocates at the end, but those deallocates don't matter anyways.

[...]

PR #94, opened 2017-05-07
Break cyclic dependency on DirtyQubitTag by moving it to projectq.tags

Strilanc:
Part of fixing issue https://github.com/ProjectQ-Framework/ProjectQ/issues/93


thomashaener:
To me, DirtyQubitTag would make sense in either meta or types. I don't think having it in cengines makes sense...

Strilanc:
`DirtyQubitTag` is implicitly part of the definition `BasicEngine`'s `allocate_qubit` method. So `DirtyQubitTag` is forced to be a dependency of `cengines`. Since `cengines` is a dependency of `meta`, `DirtyQubitTag` can't go in `meta` without creating a cycle.

One alternative would be to split `meta` into two parts, one below `cengines` and one above.

thomashaener:
I realize that this introduces a cycle, but it's a meta tag and as such it belongs into meta. If I were looking for this file, `cengines` would be the last place I'd expect it to be.

Strilanc:
How about a separate directory for tags called `tags`? That would separate the no-dependency tag classes from the many-dependency command modifier classes.

thomashaener:
Is the cyclic dependency creating issues? Adding a separate directory just for tags seems a bit extreme.

Strilanc:
In general, cyclic dependencies between modules are bad. This particular cyclic dependency is bad because it forces an import inside a method that can run during shutdown. It triggers additional errors.

Strilanc:
Is there anything holding this up?

PR #95, opened 2017-05-07
Prevent Qubit.__del__ from running more than once

Strilanc:
Currently, when the call to `engine.deallocate_qubit` fails, the `self.id = -1` line never runs. The failure can also resurrect the qubit object instance, so its del method gets scheduled to run again, and so an infinite error text loop closes.

Alternatively, the loop can be closed by "helpful" code triggered by the dealloc accidentally calling `__del__` on the qubit.

I've seen it happen (I injected a failure-on-tenth-iteration into LocalOptimizer than ran the Shor example). And this change fixed the problem.

### DIFF
--- a/projectq/types/_qubit.py
+++ b/projectq/types/_qubit.py
@@ -116,10 +116,17 @@ class Qubit(BasicQubit):
     Thus the qubit is not copyable; only returns a reference to the same
     object.
     """
+    def __init__(self, engine, idx):
+        BasicQubit.__init__(self, engine, idx)
+        self._has_run_del = False
+
     def __del__(self):
         """
         Destroy the qubit and deallocate it (automatically).
         """
+        if self._has_run_del:
+            return
+        self._has_run_del = True
         self.engine.deallocate_qubit(self)
         self.id = -1
 
--- a/projectq/types/_qubit.py
+++ b/projectq/types/_qubit.py
@@ -118,17 +118,16 @@ class Qubit(BasicQubit):
     """
     def __init__(self, engine, idx):
         BasicQubit.__init__(self, engine, idx)
-        self._has_run_del = False
 
     def __del__(self):
         """
         Destroy the qubit and deallocate it (automatically).
         """
-        if self._has_run_del:
+        if self.id == -1:
             return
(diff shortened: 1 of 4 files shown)


projectq/types/_qubit.py
```diff
@@ -116,12 +116,18 @@ class Qubit(BasicQubit):
     Thus the qubit is not copyable; only returns a reference to the same
     object.
     """
+    def __init__(self, engine, idx):
```
thomashaener: I don't think we'll need __init__ anymore.

projectq/types/_qubit.py
```diff
@@ -116,12 +116,18 @@ class Qubit(BasicQubit):
...
+        BasicQubit.__init__(self, engine, idx)
+
     def __del__(self):
         """
         Destroy the qubit and deallocate it (automatically).
         """
-        self.engine.deallocate_qubit(self)
+        if self.id == -1:
```
thomashaener: `BasicEngine` handles this check already.

projectq/types/_qubit.py
Strilanc: Done

projectq/types/_qubit.py
Strilanc: There are already engines in the codebase with a `deallocate_qubit` method that don't do this check (in Simulator and in a mock test engine). This check reduces a complicated whole-code-base proof, about what's a main engine and what's a backend and how they get called, into a one-liner.

(... If it was up to me I'd make "you have to give me a valid qubit" a precondition of `engine.deallocate_qubit` and throw when it happened. Double-free's are symptoms of bugs.)

thomashaener: Either is fine with me, as long as it's only in one location.

Strilanc: Done. `deallocate_qubit` now asserts the qubit id isn't -1 as a precondition, and `__del__` satisfies that precondition by avoiding the call. People who have a bug that accidentally frees qubit twice will be appropriately warned by the exception.

thomashaener: Thanks!

PR #96, opened 2017-05-07
Extract _uncompute_assuming_no_deallocs from ComputeEngine.run_uncompute

Strilanc:
The `run_uncompute` function is *huge*. This extracts one of its pieces.

### DIFF
--- a/projectq/meta/_compute.py
+++ b/projectq/meta/_compute.py
@@ -87,6 +87,33 @@ class ComputeEngine(BasicEngine):
         cmd.tags.append(UncomputeTag())
         return cmd
 
+    def _undo_allocate_qubit_by_id(self, qubit_id):
+        """
+        Deallocates the qubit with the given id in the main engine.
+        Args:
+            qubit_id (int):
+        """
+        matches_in_main = [q
+                           for q in self.main_engine.active_qubits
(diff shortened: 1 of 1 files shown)


Strilanc:
In the future please give a "needs changes" review instead of closing the PR, so that there isn't a disconnect between old and new when I push fixes for the comments.

projectq/meta/_compute.py
```diff
@@ -117,27 +144,9 @@ def run_uncompute(self):
         # Don't inspect each command as below -> faster uncompute
         # Just find qubits which have been allocated and deallocate them
         if len(ids_local_to_compute) == 0:
-            for cmd in reversed(self._l):
-                if cmd.gate == Allocate:
-                    qubit_id = cmd.qubits[0][0].id
-                    # Remove this qubit from MainEngine.active_qubits and
-                    # set qubit.id to = -1 in Qubit object such that it won't
-                    # send another deallocate when it goes out of scope
```
Strilanc: Ah, I could tell the assignment and `__del__` method didn't both need to be there (since `__del__` also does the assignment to -1) but I kept the wrong one. Fixed in https://github.com/ProjectQ-Framework/ProjectQ/pull/109

PR #97, opened 2017-05-07
Extract _send_rewritten_command from ComputeEngine.run_uncompute

Strilanc:
- Dropped temporary lists because they were references not copies
- Merged id-rewritten into a single nested loop via Command.all_qubits

The `run_uncompute` function is *huge*. This extracts one of its pieces.

### DIFF
--- a/projectq/meta/_compute.py
+++ b/projectq/meta/_compute.py
@@ -87,8 +87,12 @@ class ComputeEngine(BasicEngine):
         cmd.tags.append(UncomputeTag())
         return cmd
 
-    def _send_rewritten_command(self, cmd, rewrite_id_map):
+    def _send_inverse_with_rewritten_qubits(self, cmd, rewrite_id_map):
         """
+        Rewrites the given command so it uses global qubit ids instead of
+        local-to-compute-block qubit ids, then forwards the command's inverse
+        to the next engine.
+
         Args:
             cmd (projectq.ops.Command):
             rewrite_id_map (dict[int, int]):
@@ -206,7 +210,9 @@ class ComputeEngine(BasicEngine):
                     self.send([self._add_uncompute_tag(cmd.get_inverse())])
 
             else:
-                self._send_rewritten_command(cmd, new_local_id)
+                # Forward the inverse command, but with local-to-compute-block
+                # qubit ids changed to the corresponding global qubit ids.
+                self._send_inverse_with_rewritten_qubits(cmd, new_local_id)
 
     def end_compute(self):
         """
--- a/projectq/meta/_compute.py
+++ b/projectq/meta/_compute.py
@@ -87,6 +87,21 @@ class ComputeEngine(BasicEngine):
         cmd.tags.append(UncomputeTag())
         return cmd
 
+    def _send_rewritten_command(self, cmd, rewrite_id_map):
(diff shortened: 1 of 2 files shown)


projectq/meta/_compute.py
```diff
@@ -87,6 +87,21 @@ def _add_uncompute_tag(self, cmd):
         cmd.tags.append(UncomputeTag())
         return cmd
 
+    def _send_rewritten_command(self, cmd, rewrite_id_map):
```
damiansteiger: Why is this part of the old (arguable long) function moved into a private function?
It is only called at one place. This only adds another function call overhead and for understanding run_uncompute, I have to scroll up to check what this function does...

projectq/meta/_compute.py
```diff
@@ -87,6 +87,21 @@ def _add_uncompute_tag(self, cmd):
...
+            cmd (projectq.ops.Command):
+            rewrite_id_map (dict[int, int]):
+        """
+
+        if rewrite_id_map:
+            for qureg in cmd.all_qubits:
+                for qubit in qureg:
+                    if qubit.id in rewrite_id_map:
+                        qubit.id = rewrite_id_map[qubit.id]
+
```
damiansteiger: 👍  much more compact

projectq/meta/_compute.py
```diff
@@ -191,32 +206,7 @@ def add_uncompute(command, old_tags=deepcopy(cmd.tags)):
...
 
             else:
-                if len(new_local_id) == 0:
-                    # No local qubits are active currently -> do standard
-                    # uncompute
-                    self.send([self._add_uncompute_tag(cmd.get_inverse())])
-                else:
-                    # Process commands by replacing each local qubit from
-                    # compute section with new local qubit from the uncompute
-                    # section
```
damiansteiger: I would suggest to keep a similar comment in order that to understand what this `else` branch is supposed to do

projectq/meta/_compute.py
Strilanc: Extracting small private functions is a very common and widely recommended refactoring.

Extracting functions divides the problem into manageable pieces. It limits the amount of shared context between code in the main body. It reduces various measures of complexity, such as cyclomatic complexity. It allows the possibility of separately unit-testing the behavior of the small function. It creates function names that play a role similar to comments.

In my experience interviewing candidates, whether or not they approach a sub-problem by inlining it into their code or by delegating to a function is a strong predictor of whether they will finish on time.

For an online discussion of function lengths, see the software-se question ["What is the ideal length of a method for you?"](https://softwareengineering.stackexchange.com/questions/133404/what-is-the-ideal-length-of-a-method-for-you).

Strilanc: (The reason I only moved one branch is because I want to keep the PRs small. I intend to extract several other private functions out of this large method.)

projectq/meta/_compute.py
Strilanc: Done.

Strilanc:
I think this should stay in a separate function. If it's inside run_uncompute we'll have *six* levels of `for/if/if/for/for/if` indentation.

When you're inside a case inside a loop, you don't start more loops; you call for help.

PR #98, opened 2017-05-07
Rename Command.all_qubits -> all_quregs

Strilanc:
There are many other cases where qubit should be qureg, but this one was easily fixed by search-and-replace.

Note that this will merge-conflict with other pending PRs.

### DIFF
--- a/projectq/backends/_circuits/_drawer.py
+++ b/projectq/backends/_circuits/_drawer.py
@@ -233,7 +233,7 @@ class CircuitDrawer(BasicEngine):
                     m = int(m)
                     self.main_engine.set_measurement_result(qubit, m)
 
-        all_lines = [qb.id for qr in cmd.all_qubits for qb in qr]
+        all_lines = [qb.id for qr in cmd.all_quregs for qb in qr]
 
         gate = cmd.gate
         lines = [qb.id for qr in cmd.qubits for qb in qr]
--- a/projectq/cengines/_optimize.py
+++ b/projectq/cengines/_optimize.py
@@ -52,7 +52,7 @@ class LocalOptimizer(BasicEngine):
         for i in range(0, min(n, len(il))):  # loop over first n operations
             # send all gates before n-qubit gate for other qubits involved
             # --> recursively call send_helper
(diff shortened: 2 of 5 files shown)


Strilanc:
Is there anything holding this up?

damiansteiger:
I don't see any difference in `all_qubits` vs `all_quregs` hence I would stay with the old one for consistency.

We have `eng.allocate_qubit()` hence we use the term "qubit" and therefore I'd say `all_qubits` is appropriate.

Strilanc:
If a method returns a Qureg, it should not have "qubits" in the name instead of "qureg". This is especially bad in a dynamic language like python! Names are very important!

The other methods aren't examples of good naming, they are other examples of things that need to be renamed. This PR is just the first one that I renamed.

damiansteiger:
I don't think the name suggests a specific return format of `all_qubits`. Hence, looking up the docs is anyway required.

Btw, it does not return a Qureg either but rather a tuple of quregs as described in the docs.

Strilanc:
@damiansteiger I edited my comment to say "list of quregs" instead of "qureg" (probably while you were reading it; sorry).

I strongly disagree about the name suggesting a specific return type. It says "qubits", so I expect back a flat collection of qubits. And even if it didn't imply that, telling me nothing instead of the actual structure is a serious naming faux-pas. The name is the most important piece of documentation about a method that there is; we shouldn't be wasting it on words that convey no information or misleading information.

maffoo:
I agree with @Strilanc that it can be confusing to use `qubit` in the name of methods and properties that in fact return `Qureg` objects. This tripped me up in the past when using `allocate_qubit` which returns a qureg, not a qubit (there's also an unfortunate asymmetry with `deallocate_qubit` which takes a qubit as the name indicates, not a qureg). I do think there's room to clean this up and make the names more consistent.

It might also help to modify the code so that the distinction between a qubit object and a qureg of length 1 is less pronounced, but one would have to think through the consequences of that very carefully, especially given python's dynamic character.

damiansteiger:
We will consider in a following update. Currently a qubit is the only important entity. A qureg is just a list of qubits...

PR #100, opened 2017-05-12
Small change to fix TypeErrors in TimeEvolution

idk3:
The existing code can cause TypeErrors (can't convert complex to float). This should fix them for line 77. Lines 127 and 130 have the same structure but should be fine, though I suspect the casts to float there aren't necessary.


idk3:
Closing - Ryan points out I'm committing to the wrong branch... Will make a new one.

PR #101, opened 2017-05-12
Fixing small TypeError bug in TimeEvolution

idk3:
Can't convert complex to float - made the pull request earlier to the wrong branch


damiansteiger:
Thanks for catching this.

Adding a small test to catch this bug in future is always a good idea. I have added it.

Issue #74, opened 2017-05-02
Package installation is not working


Strilanc:
When I run [the install commands given in the documentation](http://projectq.ch/docs/tutorials.html#detailed-instructions-and-os-specific-hints):

    python -m pip install --user projectq

or:

    sudo apt-get install build-essential python3 python3-pip
    sudo pip3 install --user projectq

The install fails with multiple distinct errors:

```
    [...]
        ImportError: No module named pybind11
    [...]
        Command [...]/bin/python -c "import setuptools, tokenize;[...] failed with error code 1 [...]
    [...]
      File "[...]/pip/basecommand.py", line 161, in main
      UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 42: ordinal not in range(128)
```

I think the first two errors are because the documentation says I need to install "build tools" but doesn't actually say what they are or how to install them. So I wasn't able to do that. (I do have `g++` installed.) But the third error looks typical of a python2 data-vs-str encoding issue.

Strilanc:
Interestingly, when I separately run `sudo pip3 install pybind11`, then `sudo pip3 install projectq` works.

Maybe it's not correctly specified as a dependency?

Strilanc:
Repro steps:

```
mkvirtualenv tmp
pip install projectq   #FAILS

pip install pybind11
pip install projectq   #WORKS
```

damiansteiger:
pybind11 is correctly included in the `requirements.txt`:

https://github.com/ProjectQ-Framework/ProjectQ/blob/master/requirements.txt

Strilanc:
Any ideas why pip install projectq would switch from failing to succeeding after installing pybind?

thomashaener:
I have no clue and I think it's really weird that pip does not install pybind despite it being a requirement.

Strilanc:
Matthew thinks it's because of line 30 in setup.py, which has to build an Extension that uses pybind11. But this is before the setup call that actually installs pybind11.

maffoo:
It looks like this could work if we could tell pip to install the required libraries _before_ building the extension module.

Strilanc:
Line 146 creates an exception but doesn't raise it?

maffoo:
[...]

PR #110, opened 2017-05-20
Added note for minimal pip version number such that dependencies are installed in the correct order

damiansteiger:


thomashaener:
Addresses the suggestion by @maffoo in #74

PR #113, opened 2017-06-06
Fix no-iteration loop running once

Strilanc:


damiansteiger:
Thanks Craig for catching this 👍 

I added a short fix such that the LoopEngine also throws away the loop body for no-iteration loops if the backend supports the loop tag.

Moreover, I have added a Type and Value Error to Loop (integer >=0).

PR #114, opened 2017-06-06
Edit tags directly instead of using CommandModifier in BasicEngine.allocate_qubit

Strilanc:
If the allocate command could somehow re-enter then this is not a no-op change, but there's no test checking for that behavior so I assume it doesn't happen?

### DIFF
--- a/projectq/cengines/_basics.py
+++ b/projectq/cengines/_basics.py
@@ -110,8 +110,9 @@ class BasicEngine(object):
         cmd = Command(self, Allocate, (qb,))
         if dirty:
             from projectq.meta import DirtyQubitTag
-            cmd.tags += [DirtyQubitTag()]
-            self.main_engine.dirty_qubits.add(qb[0].id)
+            if self.is_meta_tag_supported(DirtyQubitTag):
+                cmd.tags += [DirtyQubitTag()]
+                self.main_engine.dirty_qubits.add(qb[0].id)
         self.main_engine.active_qubits.add(qb[0])
         self.send([cmd])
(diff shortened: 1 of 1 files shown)


damiansteiger:
> If the allocate command could somehow re-enter then this is not a no-op change, but there's no test checking for that behavior so I assume it doesn't happen?

Could you explain this in more detail?

projectq/cengines/_basics.py
```diff
@@ -107,26 +107,13 @@ def allocate_qubit(self, dirty=False):
         """
         new_id = self.main_engine.get_new_qubit_id()
         qb = Qureg([Qubit(self, new_id)])
+        cmd = Command(self, Allocate, (qb,))
         if dirty:
             from projectq.meta import DirtyQubitTag
-            if self.is_meta_tag_supported(DirtyQubitTag):
```
damiansteiger: Why is this condition deleted?

Strilanc:
> Could you explain this in more detail?

If an engine further down the chain feeds a Not gate back into the top of the chain, we may end up back at this line of code while the command modifier is still is in effect. So the dirty tag would be added to the Not gate.

This PR changes that behavior, and I was pointing that out. I suspect it's probably a bug fix instead of a breaking change, but it's worth keeping in mind.

projectq/cengines/_basics.py
Strilanc: Oops. Restored.

damiansteiger:
Thanks for the change, the code is now cleaner 👍

PR #123, opened 2017-06-19
Dockerfile for ProjectQ+FermiLib

hsim13372:
I created a Dockerfile that people can use to easily install ProjectQ and FermiLib (and their dependencies) inside a Docker image. This may encourage more people to test out ProjectQ and FermiLib by avoiding installation issues.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,7 +13,7 @@ ProjectQ - An open source software framework for quantum computing
 
 ProjectQ is an open source effort for quantum computing.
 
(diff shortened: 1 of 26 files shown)


hsim13372:
Here is an example of a Docker image with FermiLib+ProjectQ with instructions: https://github.com/hsim13372/fermiq_playground

damiansteiger:
Thanks a lot, as discussed we probably put it into the FermiLib repo.

PR #125, opened 2017-06-23
check if get_expectation_value receives correct input 

damiansteiger:
Simulator might crash otherwise.

### DIFF
--- a/projectq/backends/_sim/_simulator.py
+++ b/projectq/backends/_sim/_simulator.py
@@ -133,7 +133,7 @@ class Simulator(BasicEngine):
         num_qubits = len(qureg)
         for term, _ in qubit_operator.terms.items():
             if not term == () and term[-1][0] >= num_qubits:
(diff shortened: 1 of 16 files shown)


projectq/backends/_sim/_simulator.py
```diff
@@ -125,7 +125,16 @@ def get_expectation_value(self, qubit_operator, qureg):
...
             make sure).
+
+        Raises:
+            Exception: If `qubit_operator` acts on more qubits than present in
+                the `qureg` argument.
         """
+        num_qubits = len(qureg)
+        for term, _ in qubit_operator.terms.items():
+            if not term == () and term[-1][0] >= num_qubits:
+                raise Exception("qubit_operator acts on more qubits than"
```
thomashaener: It looks like there is a space missing in the error message.

damiansteiger: Good catch 👍

PR #132, opened 2017-07-13
optimized QubitOperator subtraction

babbush:
I improved the implementation of the __isub__ and __sub__ routines in QubitOperator so that several unnecessary copies are avoided. Surprisingly, this optimization led to significantly better performance in some code that Ian and I are working on. One should not simply implement __isub__ as += - QubitOp because that causes an unnecessary copy. In cases where these operators are large, this can create problems. I also added a few spaces to some of the docs as there was inconsistent spacing.


babbush:
I might add that I basically just copied the structure of how I previously implemented __iadd__.

PR #138, opened 2017-08-15
Bugfix set_wavefunction in pysim

damiansteiger:
If the user specifies a real valued wavefunction, then the internal state of the simulator is changed to numpy float and hence it ignores the complex parts of gates but it always gave a warning if it did.
* Fixed
* Added a test to check for this bug


damiansteiger:
In all fairness, the [doc](http://projectq.readthedocs.io/en/latest/projectq.backends.html#projectq.backends.Simulator.set_wavefunction) says one should use a `list[complex]` as the wavefunction parameter in set_wavefunction, so it is technically not a bug but a nice feature that it now also accepts a real valued wavefunction without adding +0j to all entries...

PR #140, opened 2017-08-15
Decomposition of arbitrary 1 qubit gates into Ph, Rz, and Ry.

damiansteiger:

PR #142, opened 2017-08-22
Decomposition of arbitrary controlled single qubit gate

damiansteiger:
Implements the decomposition of Barenco et al.

It comes after #140. It already contains all the commits of #140 as we need some of the functions of that pull request...


damiansteiger:
This PR needs to wait for another PR fixing the decomposition rules first...

damiansteiger:
Ready for review.
It comes after #140 and #150

PR #150, opened 2017-09-06
AutoReplacer checks parent class for replacements

damiansteiger:
* If `AutoReplacer` doesn't find a decomposition rule for a gate class, it checks if there are decomposition rules for the parent classes which might work
* Important change for rules which only require a gate matrix and hence the rule works for all `BasicGate` instances

### DIFF
--- a/projectq/cengines/_replacer/_replacer.py
+++ b/projectq/cengines/_replacer/_replacer.py
@@ -133,7 +133,9 @@ class AutoReplacer(BasicEngine):
             # the gate class of the inverse gate. If nothing is found, do the 
             # same for the first parent class, etc.
             gate_mro = type(cmd.gate).mro()[:-1]
-            inverse_mro = type(get_inverse(cmd.gate)).mro()[:-1]
+            # If gate does not have an inverse it's parent classes are 
+            # DaggeredGate, BasicGate, object. Hence don't check the last two
+            inverse_mro = type(get_inverse(cmd.gate)).mro()[:-2]
             rules = self.decompositionRuleSet.decompositions
             for level in range(max(len(gate_mro), len(inverse_mro))):
                 # Check for forward rules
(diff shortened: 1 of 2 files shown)


projectq/cengines/_replacer/_replacer.py
```diff
@@ -153,9 +153,25 @@ def _process_command(self, cmd):
                 if d.check(cmd):
                     decomp_list.append(d)
 
+            # If nothing found so far, check recursively parent classes:
             if len(decomp_list) == 0:
-                raise NoGateDecompositionError("\nNo replacement found for "
-                                               + str(cmd) + "!")
+                for parent_class in type(cmd.gate).mro()[1:-1]:
```
thomashaener: I think it would be nice to combine the loop with the code above (i.e. loop over all classes returned by mro() and also check for inverse rules).

damiansteiger: It now also checks all the parent classes of the inverse gate which might only be DaggeredGate and BasicGate so the loop is a bit more complicated than initially thought.

damiansteiger:
Let me check if there is a nicer way doing it with python's method resolution order...

damiansteiger:
Deriving a class from the gate and the inverse gate class and calling the mro() should give a sensible order
https://makina-corpus.com/blog/metier/2014/python-tutorial-understanding-python-mro-class-search-path
This can be used to find the best order to check for the class names in the dictionary. But then we also need to check for each of the class names if we need to supply the gate or the inverse gate to the recognize function...

Do you want me to implement this version with mro() or keep this one for the moment?

thomashaener:
Whichever one is easier to implement. If you go for the current implementation you can just remove object and BasicGate from the inverse mro.

damiansteiger:
Yes, that is easier and already done

PR #151, opened 2017-09-06
Decompose multi-controlled gates into single controlled gates

damiansteiger:
* Decomposes multi-controlled gates into single controlled gates
* Uses (n-1) ancilla qubits and 2 * (n-1) Toffoli gates
* Note: We should maybe later separate this rule from the others. It should only be used for hardware, e.g. the IBM setup. There it would make sense to have a first `FilterEngine` which sorts out all gates which have more than 1 control qubit and have an `AutoReplacer` in front of it with this rule (and hopefully soon some more) before an AutoReplacer works on the other rules...

This PR comes after #150

### DIFF
--- a/projectq/setups/decompositions/cnu2toffoliandcu_test.py
+++ b/projectq/setups/decompositions/cnu2toffoliandcu_test.py
@@ -79,12 +79,15 @@ def test_decomposition():
     for basis_state_index in range(0,16):
         basis_state = [0] * 16
         basis_state[basis_state_index] = 1.
-        correct_eng = MainEngine(backend=Simulator(), engine_list=[])
-
+        correct_dummy_eng = DummyEngine(save_commands=True)
+        correct_eng = MainEngine(backend=Simulator(), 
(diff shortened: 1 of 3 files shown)


projectq/setups/decompositions/cnu2toffoliandcu_test.py
```diff
@@ -0,0 +1,125 @@
...
+
+        with Control(correct_eng, correct_ctrl_qureg[:2]):
+            Rx(0.4) | correct_qb
+        with Control(correct_eng, correct_ctrl_qureg):
+            Ry(0.6) | correct_qb
+
+        test_eng.flush()
+        correct_eng.flush()
+
+        for fstate in ['0000', '0001', '0010', '0011', '0100', '0101', '0110',
```
thomashaener: Why not `for fstate in range(16)` and then pass `bin(fstate)[2:]` to get_amplitude?

damiansteiger: 👍  Done.

PR #181, opened 2017-10-24
Add support for circuits of more than five qubits

decodoku:
Lines 134 and 197 of the previous version assumed a 5 qubit circuit. They therefore caused problems when a circuit on ibmqx5 was attempted with more than 5 qubits. The replacement lines make appropriate changes for ibmqx5.


thomashaener:
The IBMCNOTMapper is for the 5-qubit chip exclusively, since we do not yet automatically map circuits for the 16-qubit chip.
When running on ibmqx5, users should import projectq.setups.ibm16 instead of the regular ibm setup. This will then load a `ManualMapper` instead of this CNOT mapper, see https://github.com/ProjectQ-Framework/ProjectQ/blob/develop/examples/ibm16.py for an example.

decodoku:
Oops. I didn't notice that. Much better than my hack.

PR #185, opened 2017-12-11
new square root X gate gate and square root Swap gate

GiggleLiu:
sqrt(X) gate, sqrt(Swap) gates and their inverse are likely frequently used gates:
[https://en.wikipedia.org/wiki/Quantum_gate](https://en.wikipedia.org/wiki/Quantum_gate)

In my recent project on quantum circuit optimization, I came across these gates.

Also, I added a new example of quantum gate zoo by stacking different gates.


GiggleLiu:
It seems pytest and python version conflict blocks the continuous-integration check.

damiansteiger:
Hi Leo,

Thanks a lot for your first contribution :)

Seems like pytest does not support python 3.3 anymore. I guess it is time to drop our python 3.3 support and then the tests should run fine.

projectq/ops/_gates.py
```diff
@@ -117,6 +117,22 @@ def __str__(self):
...
+        return 0.5 * np.matrix([[1+1j, 1-1j], [1-1j, 1+1j]])
+
+    def tex_str(self):
+        return r'$\sqrt{\neg}$'
+
+    def __str__(self):
+        return "SqX"
+
+SqX = SqXGate()
+SqXdag = SqXdagger = get_inverse(SqX)
```
damiansteiger: I haven't seen many uses of this gate which would justify an own gate name.
get_inverse or with Dagger(eng) might be easy enough in case it is ever needed

projectq/ops/_gates.py
```diff
@@ -137,6 +153,26 @@ def matrix(self):
 Swap = SwapGate()
 
 
+class SqSwapGate(BasicMathGate):
```
damiansteiger: Same two comments as for SqX

projectq/ops/_gates.py
```diff
@@ -117,6 +117,22 @@ def __str__(self):
...
+    def matrix(self):
+        return 0.5 * np.matrix([[1+1j, 1-1j], [1-1j, 1+1j]])
+
+    def tex_str(self):
+        return r'$\sqrt{\neg}$'
+
+    def __str__(self):
+        return "SqX"
+
+SqX = SqXGate()
```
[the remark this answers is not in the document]
GiggleLiu: agree

projectq/ops/_gates.py
GiggleLiu: Let's discuss more on this topic. In the future, we may need a standard reference to unify the table of gates and the naming system. 

What about this book
http://www.springer.com/cda/content/document/cda_downloaddocument/9781846288869-c2.pdf?SGWID=0-0-45-1039041-p173733620

Or, have you already got one?

damiansteiger: We tend to follow gate definition of Nielsen and Chuang as they are widely accepted in literature for the standard gates. For more exotic gates, there might be discussions.

In any case because we define the gate matrix, everyone can and should check that they use the same convention. E.g. some decompositions of ProjectQ rely on a paper by Barenco et al. in which different definitions for the basic gates have been used than in Nielsen and Chuang or in  ProjectQ. So it required some adaptations...

Daggered gates are only named if they are frequently used such as T and Tdagger but that is the only one which comes to my mind. For the moment I would suggest to not name daggered gates but let users write them by using `get_inverse()` or `with Dagger`. They can still name the gate in the user code if they like... And should a particular inverse gate be used a lot in the community, we can add the name very quickly.

GiggleLiu: As your suggested, I will remove sqXdag, and rename SqX to SqrtX.

btw, one of my reasons to add these two square root gates is they have their own widely accepted symbols. So the automatically generated Tikz graph can look better. This is an aesthetic criteria :)

damiansteiger: 👍  Thanks.
I agree, SqrtX and SqrtSwap are widely used and I would have had to implement them soon myself. So I am very happy that you did this work already  (including the nice Tikz graph symbols 💯 )

GiggleLiu:
I have finished the modification, you may have a review again.I have also changed the `tex_str` method of Daggered Gates, please make sure that this change is proper. @damiansteiger

projectq/ops/_basics.py
```diff
@@ -506,6 +506,9 @@ def math_function(x):
             return list(math_fun(*x))
         self._math_function = math_function
 
+    def __str__(self):
```
damiansteiger: Can you add a trivial test for this?

projectq/ops/_gates.py
```diff
@@ -117,6 +117,21 @@ def __str__(self):
...
 
 
+class SqrtXGate(BasicGate):
+    """ Square-root X gate class """
+    @property
+    def matrix(self):
+        return 0.5 * np.matrix([[1+1j, 1-1j], [1-1j, 1+1j]])
+
+    def tex_str(self):
+        return r'$\sqrt{\neg}$'
```
damiansteiger: In my opinion, instead of \neg (¬), it looks more natural to have X,
i.e. $\sqrt{X}
What do you or @thomashaener think?

projectq/ops/_gates.py
```diff
@@ -137,6 +152,25 @@ def matrix(self):
 Swap = SwapGate()
 
 
+class SqrtSwapGate(BasicMathGate):
```
damiansteiger: This should inherit from BasicGate as it is not a BasicMath gate

projectq/ops/_gates.py
```diff
@@ -137,6 +152,25 @@ def matrix(self):
...
+    """ Square-root Swap gate class """
+    def __init__(self):
+        BasicMathGate.__init__(self, lambda x, y: (y, x))
```
damiansteiger: This is a left over from the Swap gate and needs to be removed (as it is not a BasicMathGate)

projectq/ops/_gates.py
```diff
@@ -137,6 +152,25 @@ def matrix(self):
...
+        self.interchangeable_qubit_indices = [[0, 1]]
```
damiansteiger: 👍

projectq/ops/_gates_test.py
```diff
@@ -80,6 +80,14 @@ def test_t_gate():
     assert isinstance(_gates.Tdagger, type(get_inverse(gate)))
 
 
+def test_sqrtx_gate():
+    gate = _gates.SqrtXGate()
+    assert str(gate) == "SqrtX"
+    assert np.array_equal(gate.matrix * gate.matrix,
```
damiansteiger: This is a useful test 👍 
Can you also add a trivial test checking the gate matrix? As there are always two different roots, we might want to make sure we don't change convention of which one we pick

projectq/ops/_gates_test.py
```diff
@@ -91,6 +99,14 @@ def test_swap_gate():
     assert isinstance(_gates.Swap, _gates.SwapGate)
 
 
+def test_sqrtswap_gate():
+    gate = _gates.SqrtSwapGate()
+    gate_sq = _gates.SwapGate()
+    assert str(gate) == "SqrtSwap"
+    assert np.array_equal(gate.matrix * gate.matrix, gate_sq.matrix)
```
damiansteiger: This is a useful test 👍 
Can you also add a trivial test checking the gate matrix? As there are always two different roots, we might want to make sure we don't change convention of which one we pick

projectq/ops/_metagates_test.py
```diff
@@ -77,7 +77,8 @@ def test_daggered_gate_str():
 
 def test_daggered_gate_tex_str():
```
damiansteiger: Can you test an additional gate which has a tex_str attribute (e.g. Rx) to cover both branches?

projectq/backends/_circuits/_to_latex.py
```diff
@@ -370,6 +382,72 @@ def _gate_name(self, gate):
...
+                     ).format(op=op_mid, line=midpoint, pos=pos,
+                              dagger='^{{\dagger}}' if daggered else '')
+
+        # add two vertical lines to connect circled 1/2
+        gate_str += "\n\\draw ({}) edge[edgestyle] ({});".format(
+            self._op(lines[0]), op_mid)
+        gate_str += "\n\\draw ({}) edge[edgestyle] ({});".format(
+            op_mid, self._op(lines[1]))
+
+        if len(ctrl_lines) > 0:
```
damiansteiger: This part of the code gets never executed in the tests. Would be great to write a test covering this in order to keep our almost 100% test coverage :)

projectq/ops/_gates.py
GiggleLiu: :+1: I think \sqrt{X} is proper, \sqrt{\neg} is copied from wiki, which is less intuitive. In literature, they are used in equal frequency. We can insist on \sqrt{X}.

Issue #189, opened 2017-12-27
Suggestion: use gate objects as keys in resource counter


kevinsung:
The ResourceCounter class uses a dictionary called `gate_counts` whose keys are string representations of the gates. I was wondering: why not use the gates themselves, rather than string representations? I would find this useful for performing additional analysis of the gates, such as grouping together rotation gates even if they are by different angles, and other kinds of grouping. Right now it seems that the only way to do this is by reverse-engineering the string representation to obtain the original gate, which isn't too bad, but still it seems better not to process it unnecessarily in the first place.

I guess my concrete suggestion would be to have keys of the form `(G, k)` where `G` is a BasicGate and `k` is an integer representing the control count.

kevinsung:
Oh, I actually tried it, and got
>TypeError: unhashable type: 'AllocateQubitGate'
Interesting, I've never seen this error before!

thomashaener:
1) Yes, simply using the gate objects as keys won't work (dictionary keys need to be hashable)... one would have to define the `__hash__` function of all gates first, which isn't too much work.

2) The current `ResourceCounter` is just one example implementation and it would be nice to have other compiler engines for counting resources. It should be easy to adapt the existing implementation to facilitate the analysis you describe above. Also, at some point we want the resource counting to be performed in a hierarchical fashion to speed up the compilation process. E.g., only forward commands to the next engine if "similar" commands have not already been decomposed earlier; otherwise directly update the cost and discard the command.

Would you be willing to contribute to 1) or 2)?

kevinsung:
I'd be willing to do 1) and the first part of 2). Do you think I should change the existing implementation of `ResourceCounter` to use my suggestion? I don't think this internal change warrants creating a separate engine, even if I do add some additional functionality to it.

I may be willing to do the latter part of 2) later on when I have more time and am more familiar with ProjectQ. Perhaps you should create an issue for it?

To submit a contribution I just make a pull request, right? This isn't clear from the README.

thomashaener:
Awesome! Sounds good, thanks.
Yes, you can just make a PR whenever you think it's ready for review.

PR #190, opened 2018-01-04
Define hash for Gates, use Gate objects as keys in ResourceCounter

kevinsung:
Addresses #189 .

I defined the hash of a gate as the hash of its string representation. Note that if a subclass of `BasicGate` (or any class) defines its own `__eq__`, then Python 3 automatically sets its `__hash__` method to `None`, so you need to redefine its `__hash__` method.

### DIFF
--- a/projectq/backends/_resource.py
+++ b/projectq/backends/_resource.py
@@ -77,12 +77,12 @@ class ResourceCounter(BasicEngine):
         self.max_width = max(self.max_width, self._active_qubits)
 
         ctrl_cnt = get_control_count(cmd)
-        gate_name = ctrl_cnt * "C" + str(cmd.gate)
+        gate_description = (cmd.gate, ctrl_cnt)
 
         try:
-            self.gate_counts[gate_name] += 1
+            self.gate_counts[gate_description] += 1
         except KeyError:
-            self.gate_counts[gate_name] = 1
+            self.gate_counts[gate_description] = 1
 
     def __str__(self):
         """
@@ -95,8 +95,10 @@ class ResourceCounter(BasicEngine):
         """
         if len(self.gate_counts) > 0:
             gate_list = []
-            for gate, num in self.gate_counts.items():
-                gate_list.append(gate + " : " + str(num))
+            for gate_description, num in self.gate_counts.items():
+                gate, ctrl_cnt = gate_description
+                gate_name = ctrl_cnt * "C" + str(gate)
+                gate_list.append(gate_name + " : " + str(num))
             return ("\n".join(list(sorted(gate_list))) +
                     "\n\nMax. width (number of qubits) : " +
                     str(self.max_width) + ".")
--- a/projectq/ops/_basics.py
+++ b/projectq/ops/_basics.py
@@ -260,11 +260,9 @@ class BasicRotationGate(BasicGate):
 
             [CLASSNAME]([ANGLE])
         """
-        rounded_angle = round(self.angle, EQ_PRECISION - 1)
-        if (rounded_angle < EQ_TOLERANCE or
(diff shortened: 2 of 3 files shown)


thomashaener:
Thanks a lot! Could you also quickly add two small tests for the BasicPhase/RotationGate hash functions?

kevinsung:
Ok I did it. For some reason I can't get pytest working with this project; I keep getting the error
```
ImportError: cannot import name 'MappingProxyType'
```
I think the tests will pass though.

kevinsung:
Nevermind, I got pytest working. Now I'm just fixing the tests.

kevinsung:
Ok, I have discovered a potential problem with my scheme. It assumes that if `G1 == G2` evaluates to true, then `str(G1) == str(G2)` also evaluates to true, for any BasicGates `G1` and `G2`. This is not true for BasicRotation/PhaseGates, for instance, if they have angles that differ by a nonzero amount less than `EQ_TOLERANCE`. This can be fixed by redefining the `__str__` function to round the angles first, but in general we would need to enforce that `G1 == G2` imples `str(G1) == str(G2)`. Are you okay with having this be a restriction throughout ProjectQ? I think it is a sensible one.

thomashaener:
pytest: Great, I'm glad it's working now.
str & equality: Yes, I agree that enforcing this is reasonable.

Thanks!

kevinsung:
Ok, I wrote the tests, but making them pass is actually way more annoying than I expected due to the subtleties of floating point arithmetic. I'll have to think about it more later.

kevinsung:
To make things work out, I made the string representation have one less significant digit than the precision specified by `EQ_TOLERANCE`. I hope that's okay.

kevinsung:
This makes the tests pass, but on further thought this still doesn't work in complete generality... perhaps we need to ditch this idea entirely. The problem is that you can always have two values that are very close, but one gets rounded up and the other gets rounded down. I haven't yet thought of a good way to get around this.

kevinsung:
One possible solution is to always round the angle before storing it; that would remove the need for a custom `__eq__` function. Not sure if this is desirable.

kevinsung:
I made things work by modifying both `__str__` and `__eq__`. Let me know if you find these changes unacceptable.

thomashaener:
I think this is a good way of doing it, thanks a lot!

Coveralls reports that there are still 3 lines which never get called by the tests (the lines which set the angle to zero if it was close to 2pi or 4pi for the two `tex_str` methods and for `BasicPhaseGate.__eq__`). Could you add such angles to the tests as well? That would be great.

kevinsung:
A variable called `EQ_TOLERANCE` is also defined and used elsewhere (namely, ops/_qubit_operator.py). It's possible that the same concept will be used elsewhere in the future. What do you think of making it (perhaps EQ_PRECISION too) a global variable defined somewhere high up (maybe a file like `projectq/config.py`)?

thomashaener:
Yes, we might also want to do something like that at some point. But defining a tolerance specific to angles would facilitate the implementation of str, eq, ... since we could just round upon initialization and then also carry out single-qubit rotation synthesis using the same tolerance (once this is implemented).

Having one global tolerance may also not be enough, so we might even want to be able to specify the tolerance for every rotation gate instance... but we can also address this later.

kevinsung:
If we round the angle upon initialization, that would remove the need for an `__eq__` method altogether. I actually thought of that but wasn't sure if it was desirable to always store only the rounded angle. I think this is a good way too though; would you like me to implement it?

kevinsung:
I like the idea to round the angles upon initialization. I'll go ahead and implement it later today.

kevinsung:
Ok, I've done it. Let me know if there's anything else.

PR #196, opened 2018-01-18
RevKit interface: PermutationOracle

msoeken:
This contains a first oracle from the RevKit interface, `PermutationOracle`, to automatically synthesise permutations into reversible circuits.

### DIFF
--- a/docs/projectq.libs.revkit.rst
+++ b/docs/projectq.libs.revkit.rst
@@ -1,11 +1,20 @@
 revkit
 ======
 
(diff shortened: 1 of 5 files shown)


msoeken:
Coverage drops, because tests for `projectq.libs.revkit` are skipped, if the `revkit` module cannot be found. Can I disable coverage check for that module?

damiansteiger:
For the moment,  it is fine as it is not yet on the develop branch

PR #201, opened 2018-02-03
ControlFunctionOracle

msoeken:
This function will add "single-target gates", which are gates controlled by an arbitrary control function. The control function is passed as a truth table (in integer notation) or as Python function which is then tried to translate using the `dormouse` module into a truth table.


msoeken:
I added also a PhaseOracle that synthesises a phase circuit and applies to a given set of qubits.

Issue #192, opened 2018-01-09
For next release: Improve error handling and setup selection


thomashaener:
For the next release, we would like to make the following changes which may break backward compatibility

- [x] Improved error handling: Get rid of long error messages and maybe even add different warning levels to choose from (addresses #93)
- [x] Instead of importing setups, have a function, e.g., `projectq.select_setup('ibm')` (this also addresses #191 and the comment in #188)

PR #207, opened 2018-02-08
Improved error handling

damiansteiger:
* Stop further exceptions from being thrown once the first exception causes to terminate the program. This includes the case where the first exception is by the user code and not from within ProjectQ.
* Verbose flag for `MainEngine`. By default it is `False` which has the effect that the users only see a shortened stack trace if an exception occurred within ProjectQ. But this compact stack trace should contain all the necessary debug information in most cases.
* When MainEngine goes out of scope, it not only sends a Flush gate but also deallocates all active qubits first

(Addresses first part of #192)

### DIFF
--- a/examples/simulator_tutorial.ipynb
+++ b/examples/simulator_tutorial.ipynb
@@ -3,9 +3,7 @@
   {
    "cell_type": "markdown",
    "metadata": {
-    "collapsed": true,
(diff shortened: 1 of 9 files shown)


damiansteiger:
Fixes #93 and #89

PR #209, opened 2018-02-22
Improving the docs

damiansteiger:


damiansteiger:
Unfortunately, the autosummary from the last commit needs to be maintained manually but I think it is worth it and at some point we could write a script to make it automatic...

damiansteiger:
Ready for Review. 
For this purpose, I have turned on RTD for this branch:
http://projectq.readthedocs.io/en/fix-docs/index.html

PR #217, opened 2018-03-30
RevKit interface

msoeken:
I propose to merge the RevKit feature into develop. The tests are running with 100% coverage, the code is documented, and there are two examples. I am happy to add more functionality in the future. I also thought of having some simple Python-based fallback algorithms for synthesis that can be called if RevKit is not installed but do not guarantee the same quality of results.


thomashaener:
Awesome, thanks! It looks great except for a few pep8 issues (which I could fix if you like) and I was wondering if there is a chance we could change the writing to a file and then `exec_from_file` to directly running the ProjectQ code in `revkit`. E.g., have a function `revkit.run_projectq(qs)` or some other way to directly get the gates to apply without `exec_from_file`. If this requires too much work, we can also do this later.

msoeken:
Thanks for the feedback. And both good ideas. I can fix the pep8 issues and will also modify the interface to avoid writing to files. I'll push the changes within a few days.

projectq/libs/revkit/_control_function_test.py
```diff
@@ -0,0 +1,91 @@
...
+    saving_backend = DummyEngine(save_commands=True)
+    main_engine = MainEngine(backend=saving_backend,
+                             engine_list=[DummyEngine()])
+
+    qubit0 = Qubit(main_engine, 0)
+    qubit1 = Qubit(main_engine, 1)
+    qubit2 = Qubit(main_engine, 2)
+    qubit3 = Qubit(main_engine, 3)
+
+    ControlFunctionOracle(maj) | ([qubit0, qubit1, qubit2], qubit3)
```
damiansteiger: What does this test?

damiansteiger:
Would be great to incorporate this PR before the next release.

I fixed the pep8 issues.

A few things:
* Is there a way to exclude revkit from coveralls for Python3.5 and Python3.6?
* Unrelated to this code: Travis lately fails often with 
```
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install gcc-4.9 g++-4.9" failed and exited with 100 during .
```
Rerunning was so far successful in fixing this. But maybe we should check if we find a more stable solution ;)
* What else is missing besides a final code review?

msoeken:
The reimplementation somehow diverged the branches too much such that it was cleaner to open a new PR #241 and close this one.

PR #219, opened 2018-04-26
Fix teleport example in docs.

thomashaener:


thomashaener:
Okay the builds fail due to the new pip version breaking pybind11. I think we can just wait until https://github.com/pybind/pybind11/issues/1361 is fixed and the new version is available.

PR #221, opened 2018-04-27
Added the CommandLister class

zeldredge:
I added a class that makes CommandPrinter a little easier to use in an interactive environment by making it store a running list of all the commands. This might be easier for people than redirecting stdout, especially in a more complicated program where stdout might have other duties.


thomashaener:
Thanks! There already is such an engine though. It is called `DummyEngine` and is currently used a lot for testing purposes. If you initialize it, e.g., as
```
backend = DummyEngine(save_commands=True)
eng = MainEngine(backend=backend)
...
eng.flush()
```
you can then access a list of all received commands via `backend.received_commands`.

PR #226, opened 2018-06-13
Feature/mapper

damiansteiger:
* New BasicMapperEngine
* Automatic support for such BasicMappers for backends such as `Simulator`, `ClassicalSimulator`, `CommandPrinter`, and `ResourceCounter`
* Added mapper attribute to `MainEngine`
* Fix registering measurements in `ResourceCounter` and added `depth_of_dag` feature
* Pep8


projectq/backends/_printer_test.py
```diff
@@ -19,8 +19,12 @@
 import pytest
 
 from projectq import MainEngine
-from projectq.cengines import DummyEngine, InstructionFilter
-from projectq.ops import T, H, Command, NOT, Measure
+from projectq.cengines import (DummyEngine,
+                               InstructionFilter,
+                               LogicalQubitIDTag,
```
thomashaener: Can we add this somewhere else? E.g., types (qubit)?

projectq/backends/_resource.py
```diff
@@ -35,6 +38,9 @@ class ResourceCounter(BasicEngine):
             where ctrl_cnt is the number of control qubits.
         max_width (int): Maximal width (=max. number of active qubits at any
             given point).
+    Properties:
+        depth (int): Circuit depth. It is the longest path in the directed
```
thomashaener: I wouldn't call this circuit depth but "depth of DAG".

projectq/cengines/_main.py
```diff
@@ -64,7 +65,8 @@ def __init__(self, backend=None, engine_list=None, setup=None,
         Args:
             backend (BasicEngine): Backend to send the circuit to.
             engine_list (list<BasicEngine>): List of engines / backends to use
-                as compiler engines.
+                as compiler engines. Note: The engine list can currently only
+                contain at most one mapper (instance of BasicMapperEngine).
```
thomashaener: ...must not contain multiple mappers

damiansteiger: 👍

projectq/backends/_printer_test.py
damiansteiger: Went into `projectq.meta`

projectq/backends/_resource.py
damiansteiger: 👍 it's now called `depth_of_dag`

PR #237, opened 2018-07-17
Add CZ gate

damiansteiger:
Makes things more interesting with the new setups.

Decompositions for CZ gates are not yet optimized, i.e. it first compiles to CNOTs and then from CNOTS to CZ but it is a start.


damiansteiger:
This PR already contains #236

projectq/setups/decompositions/cnot2cz.py
```diff
@@ -0,0 +1,39 @@
...
+    CZ | (ctrl[0], cmd.qubits[0][0])
+    Uncompute(eng)
+
+def _recognize_cnot(cmd):
+    return get_control_count(cmd) == 1
+
+#: Decomposition rules
+all_defined_decomposition_rules = [
+    DecompositionRule(X.__class__, _decompose_cnot, _recognize_cnot)
+]
```
thomashaener: Could you add a new line here?

PR #238, opened 2018-07-18
Allow arbitrary backend qubit orderings in GridMapper

damiansteiger:
Potentially breaking changes (does not affect master branch):
* self.current_mapping is now a property (all ProjectQ mappers are adapted)

### DIFF
--- a/projectq/cengines/_twodmapper.py
+++ b/projectq/cengines/_twodmapper.py
@@ -40,7 +40,7 @@ class GridMapper(BasicMapperEngine):
     """
     Mapper to a 2-D grid graph.
 
-    Physical qubits on the grid are numbered in row-major order. E.g. for
+    Mapped qubits on the grid are numbered in row-major order. E.g. for
     3 rows and 2 columns:
 
     0 - 1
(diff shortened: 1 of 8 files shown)


projectq/cengines/_twodmapper.py
```diff
@@ -49,14 +49,18 @@ class GridMapper(BasicMapperEngine):
...
+    pass a dictionary translating from our row-major mapped ids to these
+    backend ids.
 
     Note: The algorithm sorts twice inside each column and once inside each
           row.
 
     Attributes:
         current_mapping:  Stores the mapping: key is logical qubit id, value
-                          is mapped qubit id from 0,...,self.num_qubits
+                          are backend qubit ids.
```
thomashaener: minor change: "value is backend qubit id"

projectq/cengines/_twodmapper.py
```diff
@@ -87,11 +101,32 @@ def __init__(self, num_rows, num_columns, storage=1000,
...
+                 len(set(self._mapped_ids_to_backend_ids.values())) ==
+                 self.num_qubits)):
+            raise RuntimeError("Incorrect mapped_ids_to_backend_ids parameter")
+        self._backend_ids_to_mapped_ids = dict()
+        for mapped_id, backend_id in self._mapped_ids_to_backend_ids.items():
+            self._backend_ids_to_mapped_ids[backend_id] = mapped_id
+        # As we use internally the mapped ids which are in row-major order,
+        # we have an internal current mapping which maps from logical ids to
+        # these mapped ids:
+        self._current_row_major_mapping = deepcopy(self.current_mapping)
```
thomashaener: Isn't this just `None`?

damiansteiger: Just in case someone changes the initial value of BasicEngine otherwise it will be hard to tell if something goes wrong in this mapper.

projectq/cengines/_twodmapper.py
damiansteiger: Good catch

PR #241, opened 2018-07-19
RevKit interface (II)

msoeken:
I propose to merge the RevKit feature branch into develop. The tests are running with 100% coverage, the code is documented, and there are two examples. I am happy to add more functionality in the future.

This integration is much simpler than the first approach of integrating RevKit, since it is based on a lightweight reimplementation using only the reversible synthesis algorithms necessary for the proposed functions.


projectq/libs/revkit/_control_function_test.py
```diff
@@ -0,0 +1,91 @@
...
+# run this test only if RevKit Python module can be loaded
+revkit = pytest.importorskip('revkit')
+
+
+def test_control_function_majority():
+    saving_backend = DummyEngine(save_commands=True)
+    main_engine = MainEngine(backend=saving_backend,
+                             engine_list=[DummyEngine()])
+
+    qubit0 = Qubit(main_engine, 0)
```
thomashaener: Can we change this to `qubit0 = main_engine.allocate_qubit()`? This will, in addition, send allocate gates so the command count below needs to be adapted but it is how qubits should be allocated.

thomashaener:
Thanks, it looks great now! There is just a minor comment I had concerning some of the tests where Qubit objects are explicitly constructed instead of properly allocating them from the MainEngine. Is there a reason for this? If not, I can quickly edit these files and then we can merge this into develop! :1st_place_medal:

msoeken:
I fixed the tests and am allocating from the `MainEngine` now. I think I have seen the explicit qubit construction in some other tests. No specific reason to use them.

PR #249, opened 2018-07-28
Example for phase estimation

fernandodelaiglesia:
Dear all,

I would like to add an example for phase estimation.

Thanks
Best


thomashaener:
Hi @fernandodelaiglesia,
Thanks, this would be a nice example to have!

The PR would require some significant cleanup though. Given that QPE is used so often in quantum algorithms, it should become a native gate operation in ProjectQ. This then also allows to provide multiple implementations if the output register is measured afterwards (iterative or replace iQFT by semi-classical one).

I think it would be nice if the example looked something like
```
from projectq.ops import PhaseEstimation
...
PhaseEstimation(U) | (qpe_ancillas, system)
```
where U is a parametric gate which gets t = 1, 2, 4, 8, ... as parameters by the decomposition rule for `PhaseEstimation`. What do you think?

fernandodelaiglesia:
Hi @thomashaener 

Thanks a lot for your comment. Completely agree. That would be my next objetive regarding QPE. I just think that it will take to me more time to create a native operation (I'm not an expert in python) that an example. Moreover I don't fully understand the decomposition mechanisms so I need to investigate and understand this before.

Do you think it is worth to first have the example as is in teh examples folder (cleaning the comments on testing if you prefer so) and start to work in the parametric gate?

If so we can discuss the requirements in a separate thread

Thank you

Fernando

thomashaener:
I don't think it will take much time to adapt it. You can make a QPE-gate similar to the gates you defined in your example, except that QPE takes a `U` as an argument (instead of, e.g., an angle of rotation).

Defining decomposition rules is fairly simple as well, see https://github.com/ProjectQ-Framework/ProjectQ/blob/develop/projectq/libs/math/_default_rules.py for an example. In your decomposition, you can then grab `U` from the gate object and perform the QPE as implemented in your example (just without the final measurement).

damiansteiger:
It might make sense to think a bit more about an extended syntax of a QPE gate and in general more advanced algorithms.

`PhaseEstimation(U) | (qpe_ancillas, system)`

Would be nice if `U` could something else than just a gate. E.g. the current workflow would be to create a new gate `U`, generate a decomposition rule for `U`, register the rule in the compiler, and then one can use PhaseEstimation...

thomashaener:
We can provide a wrapper class to turn a function into a gate (since we just need something that overloads the |-operator) and maybe even automatically wrap functions that are passed to the QPE gate.

fernandodelaiglesia:
I'm re-reading your comments and I realize that I misunderstood the suggestions. Please let me think on that and review the decomposition, etc.
Thanks

fernandodelaiglesia:
Hi @thomashaener and @damiansteiger 

Sorry for the delay to coming back to this... I had some busy days at work.

After some review and meditation ;-) I think we could go for a gate model as the one suggested by @thomashaener 
```
from projectq.ops import PhaseEstimation
...
PhaseEstimation(U) | (qpe_ancillas, system)

```
My idea is to mimic the time evolution schema: create a _phase_estimation.py under ops, as well as the corresponding _phase_estimation_test.py and add the PhaseEstimation entry to the ops.rst under docs.
What do you think?

What I don't see and don't understand is the idea on the decomposition rule. The Operator U should cover any Unitary operation where system qureg must be eigenstates of U ([https://en.wikipedia.org/wiki/Quantum_phase_estimation_algorithm](https://en.wikipedia.org/wiki/Quantum_phase_estimation_algorithm)) therefore I don't see how to create a general decomposition schema. I suggest to start using a general unitary U as in the example I submitted and if appropriate to evolve in the future.

In the other side, for the QPE algorithm it is required to know the amount of ancillas n in order to apply the Controlled U**(2**l) with l=0...n-1. How can I get this number "n" from the qpe_ancillas qureg? I was unable to find it

```
qpe_ancillas = eng.allocate_qureg(n)
PhaseEstimation(U) | (qpe_ancillas, system)
```
Should it be included in the gate as
```
qpe_ancillas = eng.allocate_qureg(n)
PhaseEstimation(U,n) | (qpe_ancillas, system)
```
?

Thanks

damiansteiger:
Hi @fernandodelaiglesia 

No worries regarding the delay. It will also take a few days on our side to propose a nice interface.

> The Operator U should cover any Unitary operation where system qureg must be eigenstates of U 

Yes, the basic idea is that we can perform phase estimation for an arbitrary `U`. I am a bit confused why you think that the system qureg must be an eigenstate. The beauty of phase estimation is that afterwards you collaps to an eigenstate. Also phase estimation does not necessarily mean you want to measure the phase if it is used in a larger algorithm.

>How can I get this number "n" from the qpe_ancillas qureg?
len(qpe_ancillas)


Things we should consider before writing code:

1) How to deal with the different types of phase estimation

2) ```PhaseEstimation(U) | (qpe_ancillas, system_qubits)``` is a nice syntax and certainly works if U is a gate (instance of `BasicGate`). However, often `U` is subprogram consisting of many qubits, i.e.:

```
def U(system_qubits):
   Rx(0.4) | system_qubits[2]
   QFT | sytem_qubits
   ...
```
here it is done as a function and some sort of similar syntax should also work for specifying `U` (I have a few proposal I will look at in the next few days). Currently the user could create a custom gate and register a decomposition rule and then use it. However, this process is too complicated and in other cases we might need something similar, so we can extend the MainEngine to package such functions into gates... More on that later.

3) Allow that U can be specialized for different powers otherwise certain algorithms won't be efficient.

fernandodelaiglesia:
Hi @damiansteiger 

Some responses and some questions:

> I am a bit confused why you think that the system qureg must be an eigenstate.

Mmmm I'm working on the Phase Estimation (PE) as defined by the wikipedia I pointed in the previous message or in the Nielsen and Chuang great book. In these, two key points are that U is not an arbitrary operation/matrix, but any Unitary matrix, and that |u> (system qubits) is a eigenvector (or eigenstate) of U. May be the "eigenstate" nomenclature was misleading. 

What you collapse are the ancillas to a eignestate of the standard base that in fact "is" the phase you are looking for in binary, but you don't measure the system_qureg

And the questions:

> How to deal with the different types of phase estimation

What do you mean by "different types of phase estimation"? As I said, I'm just going for the one defined in the references before

> often U is subprogram consisting of many qubits,

Good point. But as stated before, U must be an Unitary matrix, in general could be described as an instance of a BasicGate. I don't see U as a subprogram that acts on a qureg, but the definition of the gate itself, that will act in an especial way on the system qureg and de ancilla qureg when applying the PE Gate (still, I don't fully like the concept of PE Gate versus a PE Operation/algorithm `PhaseEstimation(U, system_qureg, number_ancillas)` where the integer number_ancillas determines the precision on the estimation and the probability of finding the exact phase (refer again to Nielsen and Chuang))

Of course the user could create the matrix U form any mechanism and including some decomposition rule. A very interesting example to be used with PE would be to have a matrix/gate as a TimeEvolution(t=time, hamiltonian=hamiltonian)

> Allow that U can be specialized for different powers otherwise certain algorithms won't be efficient

I'm sorry but I don't understand this. In the PE algorithm you need to use (and apply to system_qureg) several powers of U: U**(2**l) for l=0,1,2,...number_ancillas -1 For sure an efficient way of creating such powers is required, my proposal is to start with the simple approximation (as in my example pull request) and optimize afterwards.

Please find below the wikipedia image for PE algorithm, I think it clarifies a lot 

image

So sorry for the long message

damiansteiger:
A few fast answer now and the rest next week:

> Mmmm I'm working on the Phase Estimation (PE) as defined by the wikipedia I pointed in the previous message or in the Nielsen and Chuang great book. In these, two key points are that U is not an arbitrary operation/matrix, but any Unitary matrix, and that |u> (system qubits) is a eigenvector (or eigenstate) of U. May be the "eigenstate" nomenclature was misleading.

Sorry, I wasn't aware that wikipedia doesn't mention how phase estimation is used in quantum algorithm: So let me give you a short summary:
The input to phase estimation is a unitary operation U. The question I was raising are the different ways to specify the unitary U, it could be a basic single qubit gate or a sequence of unitary operations (which combined is also a unitary as my function above), e.g. the operator W defined in one of my papers: https://arxiv.org/abs/1711.11025 on which one performs phase estimation.

Suppose an eigenstate of U is Psi0 with eigenvalue e^(i*E0)

Wikipedia shows that if the input to phase estimation with U is the state Psi0 in the system qubits, then  the ancilla qubits after the inverse QFT encode the phase of the eigenvalue of U, i.e. E0. 

But one can also input an arbitrary state S instead of Psi0. Without loss of generality, you can write the state S as a linear combination of eigenstates of U. Simple example, 
S = 1/sqrt(2) Psi0 + 1/sqrt(2) Psi1
where Psi1 is another eigenstate of U with eigenvalue e^(i*E1).
(We don't know Psi0 or Psi1 but we know that we can write it as a superposition of eigenstates)

You can now still do phase estimation with input S because of linearity. With probability
|<Psi0|S>|^2 = 0.5 (called the overlap) you will collapse the ancillas to encode E0 and then the system qubits are in an eigenstate of U, i.e. Psi0, not before. 

So the beauty of phase estimation is that you don't need to know the eigenstate of U or the eigenvalues. Just prepare input a random state S and the output is one eigenvalue and one eigenvector (every time maybe a different one). If you are particularly interested in a specifc eigenstate, then just enter a state S which has a large overlap and hence the probability to collapse to that eigenstate is larger.

damiansteiger:
>What do you mean by "different types of phase estimation"? As I said, I'm just going for the one defined in the references before

It is fine if you are going for the phase estimation from your references. We should just make sure that there is a nice interface to choose which phase estimation one wants to make.

For example there is the semi-classical phase estimation. You can get it by moving the final measurements in your figure as far to the left as possible (controls and measurements can be swapped and exchange with classical control, see Nielsen and Chuang). Suppose the phase estimation in your figure uses n ancilla qubits, this allows to do phase estimation with only one ancilla qubit (but it is reset n times and you still need to apply the same U, U^2, ...).
It is implemented in our Shor's algorithm example:
https://github.com/ProjectQ-Framework/ProjectQ/blob/develop/examples/shor.py#L52

fernandodelaiglesia:
> A few fast answer now and the rest next week:

Completely agree ;-)

> But one can also input an arbitrary state S instead of Psi0. Without loss of generality, you can write the state S as a linear combination of eigenstates of U. Simple example,
> S = 1/sqrt(2) Psi0 + 1/sqrt(2) Psi1
> where Psi1 is another eigenstate of U with eigenvalue e^(i*E1).
> (We don't know Psi0 or Psi1 but we know that we can write it as a superposition of eigenstates)

That sound great. Agree on that as well. 

I understand that the definition of U (how it is created and decomposed) is something external to PE, and U is an input to PE.

> So the beauty of phase estimation is that you don't need to know the eigenstate of U or the eigenvalues. Just prepare as an input a random state S and the output is one eigenvalue and one eigenvector (every time maybe a different one). If you are particularly interested in a specifc eigenstate, then just enter a state S which has a large overlap and hence the probability to collapse to that eigenstate is larger.

As said before, agree. I understand that we can go for something as

```
PhaseEstimation(U) | (qpe_ancillas, system_qubits)
get_inverse(QFT) | qpe_ancillas ==> or any other iQFT as semiclasical
All(Measure) | qpe_ancillas ==> phase_i

system_qubits ===> Psi_i eigen vector for U
```
Regardig the usage of one ancilla and repeat PE for n times with U = U**(2**j) I need to understand the process. What is the adventage of doing this versus use n ancillas?

fernandodelaiglesia:
Hi,

Just lo let you know that I started to work on this taking into account the last comments. I'll let you know when I have something to pull request.

Best

fernandodelaiglesia:
I'll close this pull request to create a new one with the phase estimation gate

PR #256, opened 2018-08-13
Check for more user errors (normalized state) and extend collapse_wavefunction input

damiansteiger:
* Allow string input for `collapse_wavefunction` method of the `Simulator`
* Check that StatePreparation `final_state` is also normalized as required in the docs


damiansteiger:
Ignores now the PendingDeprecation warning of numpy that we shouldn't use numpy.matrix. We can deal with that in a few months.

PR #260, opened 2018-08-18
Phase Estimation as a Gate in ops

fernandodelaiglesia:
Dears

As commented in  #249 I created a new pull request including Phase Estimation as a new Gate in ops.

@thomashaener , @damiansteiger 
Thanks


fernandodelaiglesia:
No reviews, no comments? 😢

fernandodelaiglesia:
@thomashaener did you have the opportunity to go over ths pull? Thanks a lot in advance

fernandodelaiglesia:
Hi

Thank you for the update. I understand that I should wait for a complete plan.

Thank you
Best

thomashaener:
Sorry for the long wait!
As a first version, I think it would suffice to only support coherent QPE and (for now) ignore other versions. It would be great if you could

- Remove the custom implementation of `__or__` in the gate
- Move this implementation into a decomposition rule, including the final iQFT (see the setups folder for example decomposition rules)
- Support functions as well as gates as input: If it's a function, assume that it takes the time to evolve for and the quantum register as input. This allows to perform phase estimation efficiently for, e.g., modular multiplication, where one does not simply call the original unitary 2^k times.
- Use `with Control(...)` instead of `C(..)` so it works for both cases (gate and function)
- Clean up the comments (different indentations), run pep8

Thanks a lot and please let me know if you have any questions.

fernandodelaiglesia:
Thanks a lot for the review and sorry for the delay in the answer.
I will make the proposed changes as soon as I can (I'm having some busy weeks at work).

If I understand correctly, the issue with __or__, the one with Control and the comments questions should be easy. I need to go over the references for the decomposition rules question, that I don't undestand, as well as for the fact to support Gates and functions. Let me review the documentation, etc and I'll be back to you with my questions.

Thanks

fernandodelaiglesia:
Hi @thomashaener ,

After some review let me check if I understood the proposal:

- add PE as a decomposition rule inside setups/decompositions, this PE should include the iQFT. That would need a major refactor (no problem, once I fully understand the decomposition mechanism, that I have to re-study some more times)
  - in this use the `with Control(...)` instead of `C(...)`
- remove PE as a gate? maintain PE as a gate but including iQFT and removing de `__or__`?
- in any case run pep8 for the style guide check

I'm sorry but I don't understand the proposal to support functions. Can you please elaborate and provide some reference?

Thank you very much in advance

thomashaener:
Hi @fernandodelaiglesia 

That's right. You can keep the phase estimation gate. Just move the implementation of the phase estimation to the decomposition rule and remove `__or__`.

Support for functions: Currently, your implementation assumes that it's given a gate (on which it then performs phase estimation). It would be nice if one could also pass it a function which takes the qubits and the "time" as an argument. So if I have a function
```
def time_evolve(wavefunction, time):
    ...
```
then I would like to be able to use phase estimation without having to define a gate first (to support the `__or__`-operator syntax that your implementation currently assumes).

thomashaener:
Hi @fernandodelaiglesia,
Great, sounds good!
Regarding 1): This is expected. ProjectQ uses `projectq.setups.default` as a default engine list. If you want your decomposition to be used by default, make sure it is in `projectq.setups.decompositions`, which are the ones used by the default setup.

Regarding 2): It would be `QPE(time_evolve) | (ancillas, system_qubits)` and, in the decomposition rule, the `time_evolve` function that was passed to the `QPE`, denoted by `QPE_function` below, would be called with `system_qubits` and the `time` as arguments:
```
...
    with Control(eng, ctrl_qubits[i]):
        QPE_function(system_qubits, time=2**i)
...
```

Thanks, same to you!

fernandodelaiglesia:
Hi

Thank you for the quick answer.

Regarding 1: OK, great, thanks. I forgot to include the phaseestimation in the `__init__.py` in decompositions. Now it works great.

As for 2: If I understand correctly, due to the fact that the actions to be executed in the QPE are different if we use a Unitary operator or a function, I should include in the implementation something like:

```
if cmd.gate.unitary is a function:
   ...
   with Control(eng, ctrl_qubits[i]):
        QPE_function(system_qubits, time=2**i)
   ...
else:
   ...
   with Control(eng,ctrl_qubits[i]):
      QPE_function**(2**i)) | system_qubits
   ...

```

Correct?
Thanks in advance
Best

thomashaener:
Yes, you will have to distinguish the two cases somewhere. Please note that there is no native support for `gate ** power`, but you can use `with Loop(eng, num)` instead.

fernandodelaiglesia:
Great! Thank you.

As for the `gate**power`, yes, I know, it was just to illustrate the "if" in a simple way

fernandodelaiglesia:
Hi @thomashaener 

I commited 2 days ago the new version, but the tests seems to stuck in the travis-ci and I  cannot see the reason. Can you please help on this?

Thanks a lot

thomashaener:
Unfortunately, there is not much I can do about this; I would just try adding a new commit: Could you merge develop into your branch, resolve the merge conflict and push the commit?

projectq/ops/_qpegate.py
```diff
@@ -0,0 +1,29 @@
...
+    Quantum Phase Estimation gate.
+
+    See setups.decompositions for the complete implementation
+    """
+    def __init__(self, unitary):
+        BasicGate.__init__(self)
+        self.unitary = unitary
+
+    def __str__(self):
+        return "QPE_"+str(self.unitary)
```
thomashaener: I suggest `'QPE({})'.format(str(self.unitary))`

projectq/setups/decompositions/phaseestimation.py
```diff
@@ -0,0 +1,97 @@
...
+The unitary operator for which the phase estimation is estimated (unitary)
+is the gate in Command
+
+Example:
+    .. code-block:: python
+
+       n_qpe_ancillas = 5
+       qpe_ancillas = eng.allocate_qureg(n_qpe_ancillas)
+       system_qubits = eng.allocate_qureg(2)
+       U = unitary_specfic_to_the_problem()
```
thomashaener: Would be nice if this code could be run directly. Maybe with `U=Rz(...)` or something similar?

projectq/setups/decompositions/phaseestimation_test.py
```diff
@@ -0,0 +1,231 @@
...
+from projectq.ops import (BasicGate)
+
+from projectq.ops import QPE
+from projectq.setups.decompositions import phaseestimation as pe
+from projectq.setups.decompositions import qft2crandhadamard as dqft
+import projectq.setups.decompositions.stateprep2cnot as stateprep2cnot
+import projectq.setups.decompositions.uniformlycontrolledr2cnot as ucr2cnot
+
+
+class PhaseX(BasicGate):
```
thomashaener: These new gates should go into a separate PR.
This PR adds phase estimation only.

projectq/ops/_qpegate.py
fernandodelaiglesia: Thanks a lot for taking the time to review

Thanks for the suggestion. I'll do it

projectq/setups/decompositions/phaseestimation.py
fernandodelaiglesia: Sure, thanks.

projectq/setups/decompositions/phaseestimation_test.py
fernandodelaiglesia: I created these gates just for testing purposes, with no intention to be part of the framework. This is why I included them in the _test.py. Do you prefer I to eliminate them from the test.py (and eliminate the corresponding tests) or to maintain them in the test.py?

I'll wait for your response on this before recommit

Thanks again
Best

thomashaener: Oh I see. Is there a reason why you'd like to test with these gates specifically?
I think it would be nice to use standard functionality in the tests so other users can use them to figure out how to use QPE in their own projects. Would you mind adapting the tests to use, e.g., `Rz` or `Rx`?
If the gates you have added are useful in other settings, we can also add them to `projectq.ops` in a later PR.
Thanks!

fernandodelaiglesia: The reason is to test nontrivial phases in case of PhaseX, and to test more than 1 quibit in system qubits in the case of PhaseXxX. I can easily change the PhaseX to use Ph, but in the case of PhaseXxX I'm not sure if I can easily use a standard resource. I don't think a gate like this can be useful for others outside of testing for PE.

I propose to change PhaseX to Ph and create a new PhxX for 2-qubit system-qubits just for testing purpouses without intention to add it to `projectq.ops`  later on, what do you think? 

Thank you

thomashaener: Instead of testing a 2-qubit gate directly, you can also just test QPE by providing a function which does, e.g.,
```
CNOT | (q1, q2)
Rz(theta) | q2
CNOT | (q1, q2)
```
so you can remove both gates and use standard gates for all tests.

Thanks a lot!

fernandodelaiglesia: OK, great!

I'll change to use a 2-qubit  test using something like your proposal. Let me change the test and I'll recommit again (not before the weekend :-(  )

Thanks a lot!

fernandodelaiglesia:
Hi @thomashaener 

As promised, I've committed the changes you suggested during the weekend. They are ready to review.

Thanks a lot!

fernandodelaiglesia:
Sorry, I forgot to change some parameters in the asserts for the tests and therefore always pass. Now is more accurate.

Thanks

damiansteiger:
@fernandodelaiglesia 
Could you send me an email to info@projectq.ch?
I will need your email address to send you the contributor license agreement of ProjectQ which needs to be signed online before a PR gets merged. Thanks a lot.

fernandodelaiglesia:
Hi @damiansteiger, @thomashaener  ,

Just to check if you received the signed contributor license agreement.

In the other side, I'm thinking on creating a medium blog post explaining how to use ProjectQ's QPE, and share it with my social networks (linkedin and the like) once it get merged, what do you think? 

Thanks a lot in advance
Best

PR #263, opened 2018-08-25
Allow to apply unitary qubit operators to qubits

damiansteiger:


projectq/ops/_qubit_operator.py
```diff
@@ -226,6 +251,141 @@ def isclose(self, other, rel_tol=1e-12, abs_tol=1e-12):
...
+                       one quantum register.
+            ValueError: If quantum register does not have enough qubits
+        """
+        # Check that input is only one qureg or one qubit
+        qubits = self.make_tuple_of_qureg(qubits)
+        if len(qubits) != 1:
+            raise TypeError("Only one qubit or qureg allowed.")
+        # Check that operator is unitary
+        if not len(self.terms) == 1:
+            raise TypeError("Only unitary QubitOperators can be applied to "
```
thomashaener: Only QubitOperators consisting of a single Pauli string with a coefficient of unit length can be applied.
Otherwise, people might get confused if their operator actually is unitary but consists of multiple terms.

projectq/ops/_qubit_operator.py
```diff
@@ -226,6 +251,141 @@ def isclose(self, other, rel_tol=1e-12, abs_tol=1e-12):
...
+        # Apply new gate
+        cmd = new_qubitoperator.generate_command(new_qubits)
+        apply_command(cmd)
+
+    def get_inverse(self):
+        """
+        Return the inverse gate if QubitOperator is unitary.
+
+        Raises:
+            NotInvertible: inverse is not implemented if not unitary
```
thomashaener: if there are multiple terms or the coefficient is not 1 in absolute value

projectq/ops/_qubit_operator.py
```diff
@@ -226,6 +251,141 @@ def isclose(self, other, rel_tol=1e-12, abs_tol=1e-12):
...
+        """
+        if len(self.terms) == 1:
```
thomashaener: and abs(coefficient)...

projectq/ops/_qubit_operator.py
```diff
@@ -226,6 +251,141 @@ def isclose(self, other, rel_tol=1e-12, abs_tol=1e-12):
...
+            raise NotInvertible("BasicGate: No get_inverse() implemented.")
+
+    def get_merged(self, other):
+        """
+        Return this gate merged with another gate.
+
+        Standard implementation of get_merged:
+
+        Raises:
+            NotMergeable: merging is not implemented
```
thomashaener: These docs are a bit strange ;-)

projectq/ops/_qubit_operator.py
```diff
@@ -226,6 +251,141 @@ def isclose(self, other, rel_tol=1e-12, abs_tol=1e-12):
...
+        """
+        if (isinstance(other, self.__class__) and
+                len(other.terms) == 1 and
+                len(self.terms) == 1):
+            return self * other
+
```
thomashaener: Do we still need the comments below?

projectq/ops/_qubit_operator.py
thomashaener: 1/sqrt(2) * X + 1/sqrt(2) * Z = H

damiansteiger: Great, thanks for the example

projectq/ops/_qubit_operator.py
damiansteiger: done

projectq/ops/_qubit_operator.py
damiansteiger: done

projectq/ops/_qubit_operator.py
damiansteiger: yes, changed.

projectq/ops/_qubit_operator.py
damiansteiger: hmm nope

projectq/ops/_qubit_operator.py
thomashaener: It won't raise an exception now, though...

PR #270, opened 2018-09-10
Mathgates

peter-janderks:
Hi @thomashaener and @damiansteiger,

I added some circuits to the math library.  It includes circuits to perform addition, subtraction, division and multiplication of quantum numbers. I included the papers where I found the circuits in the file _quantummath.py. Let me know if I you would like me to change anything or if there is someway I can improve the implementation.

Cheers,
Peter-Jan


thomashaener:
Thanks a lot for working on this!

Could you add more tests such that the coverage remains the same / goes up?
Another minor change I'd suggest is to not have a conditional addition. Instead, you can just specialize the decomposition rule if cmd.control_qubits is not empty.

More to come later. Please let me know if you have questions!

peter-janderks:
I added the tests a while ago but forgot to leave a message.  The conditional addition (without carry) takes three inputs, the regular addition (with carry) takes three inputs, and the conditional addition with carry takes four inputs. So I don't think I can specialize the decomposition rule by having the control_qubits optional. I could leave out both conditional additions to simplify. Let me know if you would like me to do this or something else.

thomashaener:
Hi @peter-janderks 

Sorry for the late reply. What I meant is the following:

In ProjectQ, controlled operations are formed using, e.g., `with Control(...)`. So your conditional addition circuit should be called when a user writes
```
with Control(...):
    Add | (qr1, qr2)
```
To achieve this, you can simply check for `control_qubits` in the command object that your decomposition rule receives. If there are control qubits, use the conditional addition implementation; if there are none, use the regular addition circuit.

peter-janderks:
Hi @thomashaener 

I (finally) finished working on all the gates and the inverse gates. The `add_quantum` gate now works with `control_qubits` instead of having separate `control_add_quantum` gates.
Please let me know if there are other things I should change or work on.

Takishima:
Thank you very much for all of your work so far. I had a quick look at it today and this looks really promising.

I will do a full review of this PR when I come back from holidays (August 12)

Takishima:
Still need to look over the whole thing once more, probably next week and then we can probably merge this.

peter-janderks:
Thanks! Let me know if I can help.

projectq/libs/math/_gates.py
```diff
@@ -200,19 +204,361 @@ def __init__(self, a, N):
...
+
+    def get_math_function(self, qubits):
+        n = len(qubits[0])
+
+        def math_fun(a):
+            a[1] = a[0] + a[1]
+            if len(bin(a[1])[2:]) > n:
+                a[1] = a[1] % (2**n)
+
+                # The part below was originally written by peter-janderks
```
Takishima, line 265: This part was not entirely clear to me...

Takishima:
> Thanks! Let me know if I can help.

I mostly had one part of the code that I did not fully understand so far (see comment above)

projectq/libs/math/_gates.py
Takishima: This does make things clearer to me now. I think I'll rewrite this once again to make this intention clearer in the code.

Takishima: This also means, that in the documentation string above, the carry qubit needs to be a quantum register with 2 qubits if we want to be consistent.

Takishima (approved):

PR #272, opened 2018-09-21
add decomposition of S gate into product of T gates

peendebak:
@damiansteiger Not sure whether to make the request against dev or master


damiansteiger:
Pull request should be against the develop branch.

This seems like an unusual decomposition (changing a single Clifford gate into two `T` gates). In which context will it be used for?

peendebak:
@damiansteiger I was testing some compiler engine functionality and I noticed that there are not that many decompositions of single qubit gates so I added one. I am not sure about the implications of adding the decomposition to the list of standard decompositions. My guess is that if the S gate is available on a platform this addition has very little side effects.

Takishima:
I don't see the point of this PR. I will be closing it.

PR #275, opened 2018-09-27
add decomposition of H gate

eendebakpt:
@damiansteiger Added decomposition of H gate into X and Y rotations.

Similar to #272


eendebakpt:
@damiansteiger That works for me as well. If I need my decomposition, into just `Rx` and `Ry` I will create a PR with a general `arb1qubit2rxandry`

damiansteiger:
Great. Yes `arb1qubit2rxandry` and `arb1qubit2rxandrz` are missing

PR #283, opened 2018-11-08
WIP: Qrack GPU-accelerated simulator framework as backend

WrathfulSpatula:
Over at https://github.com/vm6502q/qrack, we've been working on an open source, (OpenCL) GPU-accelerated quantum computer simulator framework, for about a year. Qrack::QUnit also tries to explicitly Schmidt-decompose its global wave function into subsystems, which can majorly improve both memory usage and speed, (per Pednault et al., https://arxiv.org/abs/1710.05867).

I've taken the liberty of wrapping our framework into an optional ProjectQ "backend." To use it, you install Qrack, and then passing the global option "--with-qracksimulator" to the ProjectQ setup.py should compile the backend wrapper with the Qrack installation.

The Qrack framework doesn't allow the application of arbitrary-qubit-count gates, more than one qubit, but I've expanded our project's API to accept arbitrary numbers of control qubits, as your ProjectQ is designed. I've also wrapped optimized, controlled addition/subtraction, multiplication, and division gates, based on your math library. I support your simulator's interface for directly accessing the wave function amplitudes of the simulator, but you should note that Qrack::QUnit Schmidt decomposes the global state into separable subsystems, so these methods in your interface only spit out the representational equivalent of a fully entangled permutation basis wave function, which is not the raw state of Qrack::QUnit.

I've included a _qracksimulator_test.py script based on your cppsim test script, and all tests pass. Your documentation also builds. Anything else I can help with, please let me know!


damiansteiger:
Hi

Thanks for this PR and your work. It will take us a few days to have a look at it and we will write our comments as soon as possible.

By quickly looking over the files, I have noticed that Qrack is under a GPL license. While this is not a problem, it will require us to decide on a legal solution because GPL prohibit us from providing an integration into ProjectQ without our code adapting the GPL license, too. However, in order to keep ProjectQ as open as possible we insist on keeping the Apache 2 license. 

A few possible solution came to my mind:
1) We move all the interface code to your repository (GPL code is allowed to use Apache code but not vice versa)
2) Qrack might want to change to LGPL which allows us to write plugin similar to the interface of Psi4 (LGPL) into FermiLib

There might be more solutions so I will get back once I have had time to look at everything. It is probably much easier to find the best solution then. I just wanted to let you know so you have time to think about the license issues as well.

WrathfulSpatula:
I completely understand that licensing compatibility is an issue. If it would help, I can move the Qrack wrapper C++ files into the Qrack repo itself, and include them as part of Qrack's installation process. I am not a legal expert, and I do not know whether this is technically satisfactory. However, if we can think of a way to maintain our respective licensing, I can implement it, and my intent is for both projects to keep their respective licenses.

Benn Bollay (https://github.com/orgs/vm6502q/people/bennbollay) and I are the original maintainers of Qrack. I would like to keep Qrack under the GPL, but I can do whatever footwork is necessary to protect your licensing, as well. Thank you.

damiansteiger:
Thanks for the fast response. I have dealt with these licenses in the past and so I will write a longer explanation soon once I have had time to look at everything.

The important point is that it is absolutely possible for you to keep your GPL license. In order for us to keep our Apache 2 license, we cannot include your GPL code or code interfacing to your GPL code into ProjectQ. But the other way around works fine.

We can put all the interfacing code into either your current GPL repo or a separate GPL repo. In that GPL repo we are allowed to import any Apache 2 licensed code from ProjectQ (the Apache 2 license only requires to keep our headers and add our notice file). Important Apache 2 code into a GPL repo is allowed an keeps that repo under a GPL license which is more restrictive. User can install this additional backend from that Repo and use it together with ProjectQ...

WrathfulSpatula:
Thank you for the fast response, as well. I think it's important that a selection of good quantum computing and simulation resources are maintained under the (strong) GPL, but it's not Benn's and my intent to prevent you folks from maintaining Apache 2.

It wouldn't be hard to move the C++ code and whatever extent of the wrapper into our base repo, for installation from our end, if that would suffice. Otherwise, I could copy my personal fork of ProjectQ under the same vm6502q organization Qrack is under (named so for our other project, which is a quantum super set of the MOS-6502 instruction set, using Qrack).

By the way, I'm trying to run fix the coverage in this branch, based on your CI scripts, but I seem to be running into errors outside of the Qrack wrapper. I think I accidentally forked from your "development" branch instead of the master, and I'm not sure if that's at the root of the issue.

In any case, please don't hesitate to ask me for any workaround that satisfies your licensing requirements, or work to improve the implementation in the PR.

bennbollay:
Hi folks - after a bit of discussion, we're going to move the qrack project over to the LGPL license. Would that satisfy your very reasonable considerations and provide a minimum-friction interface between the two components?

We'll get on this and sort out the CI issues.

damiansteiger:
Long and detail answer is coming once I know what your code does ;-) I am a bit short in time in the next days but will try my best to have a look...

Both license (LGPL and GPL) work so it is totally up to you what you like best for your code. In both cases we would need to move the interfacing code into a separat repo. So that users need to install it and then tehy could do sth like 
```
from qrack import QRackSimulator

eng = projectq.MainEngine(backend = QRackSimulator)
```
To achieve this, you can import/copy ProjectQ code into your interfacing code as GPL/LGPL allows to include Apache 2 code (and Apache 2 allows it if you keep the header and notice file so no problem)

As a general note:
We cannot have core features of ProjectQ under GPL as it wouldn't allow us to automatically install it without loosing our Apache license. But for such additional/optional back-ends, it is anyway the user who needs to install it separately so GPL or LGPL licenses work (I wouldn't ship and install GPU code with ProjectQ by default as it might increase installation problems but provide it as an optional back-end for users who want it)

damiansteiger:
Regarding the branches:

`develop` branch is where all the latest PR are integrated (in this case I think we don't even need a PR but figure out in which repo to build the interface code for Qrack)
From time to time, we release a new version by merging the `develop` branch into the `master` branch and release that version on pypi...

WrathfulSpatula:
Damian, to be clear, there's no library code or binary from Qrack in this PR. Not a file and not a binary of Qrack is in this branch. The intent was, the user separately does a "make install" with our Qrack repo, which puts headers and a library binary in default include and library folders for the user's compiler and system. (The lack of a Qrack binary installed separately by the CI is part of the reason we're failing, at the moment.)

Once Qrack is installed this way, the branch in this PR has an optional installation feature in ProjectQ, to compile a wrapper (which is not directly part of the Qrack library) that references a separately installed Qrack header and compiles with the separately installed Qrack binary. I could be wrong, but my impression was that that's hypothetically acceptable use of an LGPL compiled library in a project under a less restrictive or commercial license.

Benn and I figuring out the CI, now. We'll get this passing pytest, while we figure out licensing compatibility. Thanks again for your attention.

WrathfulSpatula:
Damian, Benn and I have officially updated our licensing to LGPL, back on the Qrack repo. There is no Qrack library code in this PR. We reference an installed header and link against an installed library, which is acceptable use of the LGPL for an optional feature under Apache 2, as I understand it.

On my local system, I'm passing pytest with the global build/install option "--with-qracksimulator". This is an optional feature that is off by default. It requires that the Qrack library is separately installed. We're making progress, but we'll figure out a way to get your CI to pass these tests. It appears that it fails because it doesn't install the Qrack headers and object code library or build with "--with-qracksimulator".

WrathfulSpatula:
We're passing CI with the global option "--with-qracksimulator", now, but I think I'm going to move where and how Qrack gets installed, in the CI.

By the way, we tested a slightly earlier release of Qrack on an AWS P3 instance, with OpenCL, a few months back:

https://qrack.readthedocs.io/en/latest/performance.html

We think we're among the fastest simulator "back ends" out there, among those publicly available. We're happy to be even faster, with your compiler!

We added arbitrarily controlled variants of our gates specifically to interface with your compiler, and a fully entangled simulator engine object should pick up roughly a factor of 2 on gate speed for each additional control, for large numbers of qubits. We try to stay agnostic to the unknowable global wave function amplitudes, but we added enough methods for you to have rough parity with your default C++ simulator.

As your API expands, we hope to support it. Our other related project, VM6502Q, supports a quantum super set of MOS 6502 assembly, and it's designed for general amplitude amplification algorithms, like Grover's search. We have a VM6502Q assembly example of Grover's search applied to a lookup table here, and you might be interested in the unitary "IndexedLDA", "IndexedADC", and "IndexedSBC" methods we use to achieve this:

https://github.com/vm6502q/examples/blob/master/grover_asm/grover_asm.s

Thanks again!

WrathfulSpatula:
I worked on this over the weekend, obviously, and I have a few things to note.

Qrack's best performance is usually with OpenCL. Benn realized we needed to fall back to the "CPU" implementation at least for your unit tests, but we forgot to include the Qrack config.h header after that, for a while, which completely turned off GPU acceleration, I think. I just fixed this.

Also, the "CPU" implementation relies on manual SIMD intrinsics, and I discovered in the process of this PR that Qrack accidentally flips some elements in its SIMD matrix multiplication. I shunted that part of the SIMD out of Qrack temporarily for a more obvious multiplication, and I added a unit test to Qrack to alert us when the amplitudes end up with reversed signs, as a result.

I put "CPU" in quotes, because the OpenCL will still run on CPU-only systems, as almost all modern CPUs support OpenCL. This is still generally faster than the "CPU" or "non-OpenCL" implementation, despite the ostensible loss of intrinsics when running kernels designed for current generation GPUs.

I think that's pretty much all I have, for the moment. I also tried to get the coverage back up on the ProjectQ branch, but take a look at the CI coverage report, whenever you get back to this. Thanks.

WrathfulSpatula:
Since we opened this PR, we've been working on an optional gate fusion layer for Qrack. Sandwiched between the explicitly Schmidt-decomposed "QUnit" and the "Schroedinger method" "QEngine," the gate fusion layer "QFusion" gives me a speed boost, in my tests, so we added into the PR, here.

WrathfulSpatula:
Damian, we've been continuing work on the Qrack repo, and I updated this branch to correspond with the current master branch of Qrack. The CI specifically fails on only Python 3.6, and it's because the interpreter is reading "\e" and "\d" at the start of LaTeX sequences in the base ProjectQ file strings as Python string escape sequences. I'll be able to dig in tonight, to repair the LaTeX/string literal confusion, but it'd be a pain if debugging strings in the base ProjectQ files led to merge conflicts, down the road.

Qrack has been LGPL for a couple of months, now, by the way. My intent with and understanding of our chosen license is that this is Apache-2.0 compatible, in this case, as an optional feature linked against our compiled library. Such a use case can be published under any choice of license, is my reading and intent.

WrathfulSpatula:
I don't want this PR to become monolithic, apologies, but I was looking the SimulaQron repo, and I realize this PR should fit into that stack. I will adjust the interface, this weekend, so that the Qrack simulator fits the existing interfaces to complete the SimulaQron, to ProjectQ, to Qrack stack as simply as possible.

WrathfulSpatula:
As per my last comment, before you consider merging this branch, you might want to look at the PR that's just about ready to roll into it, to optionally exchange the Qrack Simulator in place of the default C++ Simulator throughout, via a global build flag on the command line: https://github.com/WrathfulSpatula/ProjectQ/pull/4

WrathfulSpatula:
I have successfully constructed a 2 physical node SimulaQron -> ProjectQ -> Qrack stack, as of last night, with https://github.com/WrathfulSpatula/ProjectQ/pull/4. Beyond the usual setup of SimulaQron with ProjectQ, all that is necessary is that Qrack is installed and then that ProjectQ is built and installed with the global flag "--with-qracksimulator". With that PR, the Qrack Simulator is switched out everywhere in place of the default C++ simulator.

Since that successfully completes the stack and has been tested, I am merging in that PR. This is the right way to initially integrate the new simulator, now that I see it in context of the network -> compiler -> simulator stack.

I don't plan new development on this PR, until it is reviewed, except to keep it up-to-date with the state of the development branch.

WrathfulSpatula:
Damian, the reason for the drop in coverage is because the decomposition tests can't reconcile the difference between direct calls to the simulator and decompositions, so I temporarily skipped these tests. I think the primary issue is that Qrack floors the residue of gates that should take a nonzero amplitude to a zero amplitude, at about the 10^(-11) scale for float accuracy.

For example, say we apply two Hadamard gates to a bit: "H(0); H(0);" Ideally, this should return bit 0 to its exact original state. If bit 0 was in a pure |0> permutation state to start, a residue is left in the |1> permutation state. For Qrack, we've measured what this residue typically is, around 10^(-11) for a wide range of applications on my test systems, and we floor it after every gate. Additionally, the Qrack::QUnit type explicitly separates bit subsystems that are separable, that are without entanglement, with a conservative approach to switching to a entangled representation, whenever it could be necessary. This reduces the number of permutation amplitudes actually stored, maintaining separate subsystems, which should also reduce float rounding error.

So, Qrack is able to exactly hit permutation basis eigenstates on gate sequences where the decompositions appear to "bleed," a little bit, and this bleed is larger than a reasonable tolerance. (You could run pytest without the tests skipped by Qrack, to see what I mean, or you could look at the previous results of CI before I skipped these tests.) The issue might be partially with Qrack, and I'll continue to look into that.

I don't think it's actually reasonably possible for either simulator to hit the default 10^(-12) tolerances of the unit tests as absolute measures of error, with float accuracy by these general methods, since I'd guess that your "residue" on idempotent sequences is probably also around 10^(-11). The tests might be self-consistent, to within that tolerance, but I don't think that's the actual amortized error magnitude.

WrathfulSpatula:
For those following along, I think I got to the bottom of this: Qrack doesn't yet implement your "uniformly controlled rotation" gates. However, it doesn't seem hard to do. We're working on it.

Beyond that, we're missing a bridge between what ProjectQ does for time evolution, with what look like Pauli basis vector decompositions, and what Qrack does, in terms of controls on arbitrary single bit Hamiltonian terms. We do it differently, because we want everything in terms of arbitrarily controlled single bit gates, (excepting swap variants), but maybe there's a more sensible way for us to do it. The time evolution is a little more math, but we'll figure it out. Uniformly controlled gates will be coming in the next day or two.

WrathfulSpatula:
Per #302 I was able to make some scripts that isolate some of the remaining issues. The Gist there points to a problem while setting wave function amplitudes with a "QFusion" gate fusion layer. (Qrack was written to be probability amplitude agnostic.)

Even if this passes unit tests, it's not entirely safe until I remove the work-in-progress designation.

WrathfulSpatula:
This has been debugged, but this PR itself has admittedly become a totally unmanageable mess, months later. Apologies for the thrashing.

I'm going to squash this whole thing, over on my personal fork, and give you one succinct comment, with one tidied commit, to start. We can work from there, if Qrack integration is something you'd like to optionally support under the Apache 2.0 license. Thank you again.

PR #284, opened 2018-11-10
pytest > 3.3 does not work

bennbollay:
Lock pytest to the 3.3 version.  The current HEAD of 3.10 fails with a wide variety of string parsing and other language errors.


damiansteiger:
Sorry for the delayed response

pytest seems to work again therefore I close this PR.

PR #285, opened 2018-11-18
Use vector's resize when allocating qubit

Wojtek242:
Reallocating a vector is bottlenecked on memory speed not CPU so it's actually slightly more efficient to use the built-in resize method (unless using big machines with multiple memory channels which  as I understand are not the target for ProjectQ). 

I extracted the simulator and benchmarked using my own script, so hopefully you can reproduce an ~10-20% speed up for allocations. The speed up will probably not be noticeable overall as I assume this isn't a particular bottleneck.

### DIFF
--- a/projectq/backends/_sim/_cppkernels/simulator.hpp
+++ b/projectq/backends/_sim/_cppkernels/simulator.hpp
@@ -55,11 +55,7 @@ public:
     void allocate_qubit(unsigned id){
         if (map_.count(id) == 0){
             map_[id] = N_++;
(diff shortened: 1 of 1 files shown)


thomashaener:
The custom resize method is indeed targeting systems with multiple NUMA domains. I realize that there might be a slight slow-down due to the spawning of threads which is unnecessary if there aren't multiple domains, but this shouldn't affect the overall runtime too much.

Wojtek242:
Ah, my bad. I have misread/misinterpreted a statement I saw in you ProjectQ paper. It makes sense now.

Issue #282, opened 2018-11-02
Significant performance improvement by implementing __eq__()


cgogolin:
I recently did some simple profiling with cProfile on circuits that consist mainly of single qubit rotations and CNOTs. It turned out that in my case about half (!) of the accumulated time was spend in `BasicGate.__eq__()`. 

The reason for this is that `_send_qubit_pipeline()` does a lot of operation comparisons when finding the `gateloc`. As `CNOT` is the controlled version of `XGate()` and the latter is a `SelfInverseGate`, which is a subclass of `BasicGate`, the `__eq__()` method of `BasicGate` is called a lot. Now, the problem is that `BasicGate.__eq__()` does a comparison of the matrices of the involved gates (!) if they both have one. 

When comparing a `CNOT` or a `XGate` with another `BasicGate` with a `matrix` this is clearly extremely wasteful, as any two `XGate`s that act on the same qubit are equal. By adding the following `__eq__()` method to `XGate` 
```
    def __eq__(self, other):
        return isinstance(other, self.__class__)
```
I as able to reduce the total runtime of my circuit from 409.136 seconds to 172.695 seconds!!!

I would directly provide a pull request that adds an `__eq__()` method to `XGate` and other similar gates, but would first like to discuss this. I think, that all gates whose `matrix` property is fixed (i.e., is always the same for any two instances of the respective class) should overwrite `__eq__()`. As there are many such gates, is is maybe worth introducing a new new subclass (called e.g., `FixedMatrixGate`) that implements this more efficient `__eq__()` and then have all those gates inherit from that class. The only downside is that if some other (possibly user-defined) `BasicGate` or a `RotationGate` happens to be by chance almost equal to, say, a `XGate`, then this would no longer by recognized. This however seems to be unlikely in real world code.

damiansteiger:
That's a very good observation. We lately changed `BasicGate` to compare matrices as we (and others) have been using it to define custom gates defined by matrices for quick testing. In some applications this might lead to a performance bottleneck and we have been planning to change it in the next release as we noticed this slowdown in a different context, too. It will be a breaking change: probably by adding a special gate class for gates whose matrix is allowed to change and remove matrix comparisons for all others...

This would be a great first pull request to get started to contribute to ProjectQ :-) If you like to do it, I can write a list of changes which I would suggest to compare with your solution and then you could start to work on a pull request. How does this sound?

cgogolin:
Sounds good :-)

[...]

Issue #287, opened 2018-11-19
`PendingDeprecationWarning` for `numpy.matrix`


AckslD:
The matrix-class of `numpy` is currently being deprecated. I therefore get the following warning when running the below example using `numpy>=1.15.0`:
```
/path/projectq/ops/_gates.py:57: PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
  return 1. / cmath.sqrt(2.) * np.matrix([[1, 1], [1, -1]])
```

Example:
```
import projectQ as pQ
import unittest

 class Test(unittest.TestCase):
         def test(self):
                 eng = pQ.MainEngine()
                 q = eng.allocate_qubit()[0]
                 pQ.ops.H | q
                 pQ.ops.Measure | q
                 eng.flush()
                 print(int(q))
```

AckslD:
An option would be to use `numpy.dot` instead of `@` if you don't want to wait for Python2.7 to be deprecated.

PR #288, opened 2018-11-26
Implement MatrixGate and simplify __eq__ in BasicGate to improve performance

cgogolin:
Addresses #282 by implementing a MatrixGate base class for gates that are defined via their matrix (similar to how, e.g., RotationGate captures gates defined via an angle).

Specifically this PR does:

* Simplify `__eq__()` in `BasicGate` to only do class/type comparison
* Disallow the usage of `.matrix` in `BasisGate` altogether (this allows to no longer worry about `==` between `MatrixGates` and `BasicGates` as they are now always not equal.
* Implement `__eq__()` in `MatrixGate` that does matrix comparison (slow but this is the thing to do for gates that are really defined via their matrix)
* Adds unit tests to check various comparisons between `MatrixGate` and other gates.
* Unlocks the performance gain discussed in #282 

Please have a look and tell me what do do/don't like. Please pay especially close attention to the changes unit tests. I am not 100% sure I modified `_simulator_test.py` in the correct way.

### DIFF
--- a/projectq/backends/_sim/_simulator_test.py
+++ b/projectq/backends/_sim/_simulator_test.py
@@ -29,7 +29,7 @@ import scipy.sparse.linalg
 from projectq import MainEngine
 from projectq.cengines import (BasicEngine, BasicMapperEngine, DummyEngine,
                                LocalOptimizer, NotYetMeasuredError)
-from projectq.ops import (All, Allocate, BasicGate, BasicMathGate, CNOT,
+from projectq.ops import (All, Allocate, BasicGate, MatrixGate, BasicMathGate, CNOT,
                           Command, H, Measure, QubitOperator, Rx, Ry, Rz, S,
                           TimeEvolution, Toffoli, X, Y, Z)
 from projectq.meta import Control, Dagger, LogicalQubitIDTag
@@ -97,37 +97,32 @@ def mapper(request):
         return None
 
 
-class Mock1QubitGate(BasicGate):
-        def __init__(self):
-            BasicGate.__init__(self)
-            self.cnt = 0
+class Mock1QubitGate(MatrixGate):
+    def __init__(self):
+        MatrixGate.__init__(self)
+        self.cnt = 0
 
-        @property
-        def matrix(self):
-            self.cnt += 1
-            return [[0, 1], [1, 0]]
+    @property
+    def matrix(self):
+        self.cnt += 1
+        return [[0, 1], [1, 0]]
 
 
-class Mock6QubitGate(BasicGate):
-        def __init__(self):
-            BasicGate.__init__(self)
-            self.cnt = 0
+class Mock6QubitGate(MatrixGate):
+    def __init__(self):
+        MatrixGate.__init__(self)
+        self.cnt = 0
 
-        @property
-        def matrix(self):
-            self.cnt += 1
-            return numpy.eye(2 ** 6)
+    @property
+    def matrix(self):
+        self.cnt += 1
+        return numpy.eye(2 ** 6)
 
 
 class MockNoMatrixGate(BasicGate):
-        def __init__(self):
-            BasicGate.__init__(self)
-            self.cnt = 0
-
-        @property
-        def matrix(self):
-            self.cnt += 1
-            raise AttributeError
+    def __init__(self):
+        BasicGate.__init__(self)
+        self.cnt = 0
 
 
 def test_simulator_is_available(sim):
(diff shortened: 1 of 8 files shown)


damiansteiger:
Thanks for the PR! I will do the review in the next days.

For legal reason, we require a standard contributor license agreement (CLA) from all contributors to ProjectQ. This CLA is a standard contract used by many Apache 2 licensed projects to give us a non-exclusive license to your contributions (so that we can, for example, redistribute it here on GitHub under the Apache 2 license). It can be signed online. Could you send me an empty email to [info@projectq.ch](mailto:info@projectq.ch) so that I can send the CLA to your email for online signing?

damiansteiger:
CLA was received :) I have set aside time next week to work on ProjectQ, in particular your pull request

damiansteiger:
A short update:

I am very sorry for the delay. This PR looks fine and increases the performance significantly, thanks a lot for that 👍 

As mentioned in #282 there is one downside to this approach that 

```
MatrixGate([[0,1],[1,0]]) != XGate
```

I will do some experiments with adding a branch `isinstance(other, MatrixGate)`  to `BasicGate.__eq__ `. If the slowdown is not significant, than I would prefer that solution and otherwise your current solution. I will report back the performance asap.

projectq/backends/_sim/_simulator_test.py
```diff
@@ -29,7 +29,7 @@
 from projectq import MainEngine
 from projectq.cengines import (BasicEngine, BasicMapperEngine, DummyEngine,
                                LocalOptimizer, NotYetMeasuredError)
-from projectq.ops import (All, Allocate, BasicGate, BasicMathGate, CNOT,
+from projectq.ops import (All, Allocate, BasicGate, MatrixGate, BasicMathGate, CNOT,
```
damiansteiger: Line >80 characters and alphabetic order for inputs

damiansteiger:
Any comments on the changes which I have commit to this PR?

Otherwise it looks good to me and should be incorporated into a new pip version asap.

projectq/ops/_basics.py
```diff
@@ -262,12 +272,19 @@ def __eq__(self, other):
             return True
         return False
 
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
```
cgogolin: Why is that necessary? Isn't `__ne__()` with the same implementation inherited from `BasicGate`?

projectq/ops/_basics.py
```diff
@@ -262,12 +272,19 @@ def __eq__(self, other):
...
     def __str__(self):
-        raise NotImplementedError('This gate does not implement __str__.')
+        return("MatrixGate(" + str(self.matrix.tolist()) + ")")
```
cgogolin: Why do you convert to list here? Isn't it better if it is visible in the output that self.matrix is a numpy array (i.e., no commas between elements vs. commas between elements for `list`s)?

projectq/ops/_basics.py
```diff
@@ -262,12 +272,19 @@ def __eq__(self, other):
...
+
+    def __hash__(self):
+        return hash(str(self))
```
cgogolin: Again, this seems superflous because `__hash__()` is inherited.

projectq/ops/_basics.py
```diff
@@ -246,9 +258,7 @@ def matrix(self, matrix):
         self._matrix = np.matrix(matrix)
 
     def __eq__(self, other):
-        """ Return True if equal (i.e., instance of same class).
-        """
-        if not isinstance(other, self.__class__):
+        if not hasattr(other, 'matrix'):
```
cgogolin: I achnowledge that the docstring was not good, but do you really want no docstring at all?

projectq/ops/_basics.py
```diff
@@ -203,20 +203,15 @@ def __or__(self, qubits):
...
-        raise AttributeError("BasicGate has no matrix property. Use MatrixGate instead.")
-
-    @matrix.setter
-    def matrix(self, matrix=None):
-        raise AttributeError("You cannot set the matrix property of gates derived"
-                             "from BasicGate. Please use MatrixGate instead if"
-                             "you need that functionality.")
-
     def __eq__(self, other):
         """ Return True if equal (i.e., instance of same class).
```
cgogolin: Maybe better "Return True if instance of the same class, unless other is an instance of :class:`MatrixGate`, in which case equality is to be checked by testing for existence and (approximate) equality of matrix representations."

cgogolin:
I leave it to you to decide on and implement my requested changes. However you decide, green light for the merge.

projectq/ops/_basics.py
damiansteiger: 👍

projectq/ops/_basics.py
damiansteiger: Sorry my bad, of course it uses `__ne__` from BasicGate. 
(It's a habit from Python2's old classes)

projectq/ops/_basics.py
damiansteiger: 👍

projectq/ops/_basics.py
damiansteiger: Just so that the `CommandPrinter` prints such a gate on one line. A numpy matrix string has line breaks which then looks slightly weird to me.

We can use numpy arrays for this string once we switch from numpy matrix to numpy arrays #287

damiansteiger:
Thanks for the review. I have included your comments into this version which definitely makes the code better 👍

projectq/ops/_basics.py
damiansteiger: uups. Python3 seems to remove __hash__ if __eq__ is redefined...

Issue #279, opened 2018-10-26
StatePreparation causes "Circuit cannot be mapped without using Swaps" on IBM


cgogolin:
Consider the following rather simple program:

```
import projectq as pq
from projectq.ops import All, X, StatePreparation, Measure
from projectq import MainEngine
import projectq.setups.ibm as ibm_setup

backend = pq.backends.IBMBackend(user='XXX', password='XXX')
eng = MainEngine(backend, engine_list=ibm_setup.get_engine_list())

qureg = eng.allocate_qureg(4)
StatePreparation([0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]) | qureg
All(Measure) | qureg
eng.flush()
```
which should be realizable by an application of X operations to some of the qubits.

On the IBM backend, it results in the following error:
```
RuntimeError: Circuit cannot be mapped without using Swaps. Mapping failed.
 raised in:
'  File "/home/cgogolin/.local/lib/python3.5/site-packages/projectq/cengines/_ibm5qubitmapper.py", line 137, in _run'
'    raise RuntimeError("Circuit cannot be mapped without using "'
``` 
The same happens for any other StatePreparation of basis state, including the trivial one, which shouldn't do anything. On up to three qubits all basis state preparations work flawlessly.

After changing to the simulator by putting 
```
backend = pq.backends.Simulator()
eng = MainEngine(backend)
```
instead of the corresponding lines in the code above, everything also works on four qubits.

cgogolin:
I am closing this as #289 will provide an easy way to do what I was trying to achieve here.

Issue #27, opened 2017-03-17
Add pylint presubmit check


Strilanc:
For example, pylint would have caught https://github.com/ProjectQ-Framework/ProjectQ/issues/19 via an undefined variable warning.

damiansteiger:
Another vote to add pep8 tests to Travis #289

Takishima:
Now using pre-commit.

PR #289, opened 2018-11-28
Add a convenient state preparation for computational basis states 

cgogolin:
One thing that is great about ProjectQ is that it has many convenience functions and gates already built in. One thing that I was missing was an easy way to prepare computational basis states.

Inspired by #279 this PR adds a simple BasisState state preparation gate that can be feed an array of 0s and 1s and prepares the corresponding computational basis state  (if executed right at the beginning of the circuit, while the system is in state |0>) by applying PauliX gates to the respective qubits.

Please have a look and see if you would find that a useful addition to ProjectQ.

### DIFF
--- a/projectq/ops/_gates.py
+++ b/projectq/ops/_gates.py
@@ -46,7 +46,6 @@ from ._basics import (BasicGate,
 from ._command import apply_command
 
 
-
 class HGate(SelfInverseGate):
     """ Hadamard gate class """
     def __str__(self):
@@ -353,14 +352,13 @@ class FlipBits(SelfInverseGate):
                 FlipBits([0, 1]) | qureg
 
         Args:
-            bits_to_flip(list[int]|list[bool]|str|int): int or array of 0/1, True/False, or
-                                    string of 0/1 identifying
-                                    the qubits to flip. In case of int, the bits to flip are
-                                    determined from the binary digits, with the least significant
-                                    bit corresponding to qureg[0]. If bits_to_flip is negative,
-                                    exactly all qubits which would not be flipped for the input
-                                    -bits_to_flip-1 are flipped, i.e., bits_to_flip=-1 flips all
-                                    qubits.
+            bits_to_flip(list[int]|list[bool]|str|int): int or array of 0/1,
+               True/False, or string of 0/1 identifying the qubits to flip.
+               In case of int, the bits to flip are determined from the
+               binary digits, with the least significant bit corresponding
+               to qureg[0]. If bits_to_flip is negative, exactly all qubits
+               which would not be flipped for the input -bits_to_flip-1 are
+               flipped, i.e., bits_to_flip=-1 flips all qubits.
         """
         SelfInverseGate.__init__(self)
         if isinstance(bits_to_flip, int):
@@ -377,8 +375,9 @@ class FlipBits(SelfInverseGate):
     def __or__(self, qubits):
         quregs_tuple = self.make_tuple_of_qureg(qubits)
         if len(quregs_tuple) > 1:
-            raise ValueError(self.__str__()+' can only be applied to qubits,' \
-                             'quregs, arrays of qubits, and tuples with one individual qubit')
+            raise ValueError(self.__str__()+' can only be applied to qubits,'
+                             'quregs, arrays of qubits, and tuples with one'
+                             'individual qubit')
         for qureg in quregs_tuple:
             for i, qubit in enumerate(qureg):
                 if (self.bits_to_flip >> i) & 1:
--- a/projectq/ops/_gates.py
+++ b/projectq/ops/_gates.py
@@ -382,8 +382,7 @@ class FlipBits(SelfInverseGate):
     def __eq__(self, other):
         if isinstance(other, self.__class__):
             return self.bits_to_flip == other.bits_to_flip
-        else:
-            return False
+        return False
 
     def __hash__(self):
         return hash(self.__str__())
--- a/projectq/ops/_gates.py
+++ b/projectq/ops/_gates.py
@@ -359,7 +359,7 @@ class FlipBits(SelfInverseGate):
 
         Args:
             bits_to_flip(list[int]|list[bool]|str|int): int or array of 0/1, True/False, or
-                                    string of 0/1 of length len(qureg) identifying
+                                    string of 0/1 of length up to len(qureg) identifying
                                     the qubits to flip. In case of int, the least significant
                                     bit indicates whether qureg[0] is flipped.
         """
@@ -375,9 +375,9 @@ class FlipBits(SelfInverseGate):
 
     def __or__(self, qubits):
         for qureg in self.make_tuple_of_qureg(qubits):
-            for i, qubit in enumerate(qureg):
-                if self.bits_to_flip[i]:
-                    XGate() | qubit
+            for i, flip in enumerate(self.bits_to_flip):
+                if flip:
+                    XGate() | qureg[i]
 
     def __eq__(self, other):
         if isinstance(other, self.__class__):
--- a/projectq/ops/_gates.py
+++ b/projectq/ops/_gates.py
@@ -44,7 +44,7 @@ from ._basics import (BasicGate,
                       FastForwardingGate,
                       BasicMathGate)
 from ._command import apply_command
-from projectq.types import BasicQubit
+
 
 
 class HGate(SelfInverseGate):
@@ -362,12 +362,13 @@ class FlipBits(SelfInverseGate):
                                     -bits_to_flip-1 are flipped, i.e., bits_to_flip=-1 flips all
                                     qubits.
         """
+        SelfInverseGate.__init__(self)
         if isinstance(bits_to_flip, int):
             self.bits_to_flip = bits_to_flip
         else:
             self.bits_to_flip = 0
-            for c in reversed(list(bits_to_flip)):
-                bit = 0b1 if c == '1' or c == 1 or c == True else 0b0
+            for i in reversed(list(bits_to_flip)):
+                bit = 0b1 if i == '1' or i == 1 or i is True else 0b0
                 self.bits_to_flip = (self.bits_to_flip << 1) | bit
 
     def __str__(self):
@@ -376,7 +377,8 @@ class FlipBits(SelfInverseGate):
     def __or__(self, qubits):
         quregs_tuple = self.make_tuple_of_qureg(qubits)
         if len(quregs_tuple) > 1:
-            raise ValueError(self.__str__()+' can only be applied to qubits, quregs, arrays of qubits, and tuples with one individual qubit')
+            raise ValueError(self.__str__()+' can only be applied to qubits,' \
+                             'quregs, arrays of qubits, and tuples with one individual qubit')
         for qureg in quregs_tuple:
             for i, qubit in enumerate(qureg):
                 if (self.bits_to_flip >> i) & 1:
--- a/projectq/ops/_gates.py
+++ b/projectq/ops/_gates.py
@@ -360,13 +360,16 @@ class FlipBits(SelfInverseGate):
         Args:
             bits_to_flip(list[int]|list[bool]|str|int): int or array of 0/1, True/False, or
                                     string of 0/1 of length up to len(qureg) identifying
-                                    the qubits to flip. In case of int, the least significant
-                                    bit indicates whether qureg[0] is flipped.
+                                    the qubits to flip. In case of int, the bits to flip are
+                                    determined from the binary digits, with the least significant
+                                    bit corresponding to qureg[0]. If bits_to_flip is negative,
+                                    exactly all bits which would not be flipped for the input
+                                    -bits_to_flip are flipped.
         """
-        if isinstance(bits_to_flip, str):
+        if isinstance(bits_to_flip, int):
+            self.bits_to_flip = bits_to_flip
+        elif isinstance(bits_to_flip, str):
             self.bits_to_flip = list([ c != "0" for c in bits_to_flip])
-        elif isinstance(bits_to_flip, int):
-            self.bits_to_flip = list([ c != "0" for c in reversed(list('{0:0b}'.format(bits_to_flip)))])
         else:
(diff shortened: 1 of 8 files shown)


thomashaener:
Thanks @cgogolin for this PR!

I have a few minor suggestions:
- Change the name to something like "FlipBits", to make it clear that this gate can be used for more than just preparing basis states
- Support integers, strings, and lists of bool or [0/1] as input
- Remove the additional note in the documentation of StatePreparation: The statement is true for all high-level gates since gates on current hardware are very noisy

cgogolin:
All suggested changes implemented with the latest commits.

I am now no longer sure whether this Gate actually fits in `_state_prep.py`. Shall I rather move it to `_gates.py`?

thomashaener:
Great, thanks! Yes, `_gates.py` would be a better fit.

cgogolin:
Done. From my side this is now read to merge.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,49 @@ def get_inverse(self):
...
+                qureg = eng.allocate_qureg(2)
+                FlipBits([0, 1]) | qureg
+
+        Note:
+            The amplitude of state k is final_state[k]. When the state k is
+            written in binary notation, then qureg[0] denotes the qubit
+            whose state corresponds to the least significant bit of k.
+
+        Args:
+            bits_to_flip(list[int]|list[bool]|str): array of 0/1, True/False or
```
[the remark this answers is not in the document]
cgogolin: Good idea. Done.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,49 @@ def get_inverse(self):
...
+                                    string of 0/1 identifying the qubits to flip
+                                    of length len(qureg).
+        """
+        if isinstance(bits_to_flip, str):
+            self.bits_to_flip = list([ c != "0" for c in bits_to_flip])
+        else:
+            self.bits_to_flip = list(bits_to_flip)
+
+    def __str__(self):
+        return "FlipBits"
```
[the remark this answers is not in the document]
cgogolin: Done.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,49 @@ def get_inverse(self):
...
+    def __or__(self, qubits):
+        for qureg in self.make_tuple_of_qureg(qubits):
+            for i, qubit in enumerate(qureg):
+                if self.bits_to_flip[i]:
+                    XGate() | qubit
+
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return self.bits_to_flip == other.bits_to_flip
+        else:
```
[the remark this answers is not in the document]
cgogolin: Done. Note that now that we allow an int to specify the bits to flip, it made sense to allow arrays that are shorter than `len(qureg)`. `__eq__()` will treat gates that flip the same bits but have `self.bits_to_flip` of different length (e.i., the differ by trailing `false` values) as not equal. I think this is the right behavior as these gates can be applied to `qureg`s of different length.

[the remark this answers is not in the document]
thomashaener: Thanks! An alternative would be to store the integer instead of storing the list of bits (and then also compare integers). Furthermore, such an implementation would work as I'd expect for the input integer `-1`: It flips all qubits in the quantum register, no matter its length. What do you think?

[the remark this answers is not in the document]
cgogolin: I am not sure this is the result I would "expect" that behavior, as for negative n, n-1, and not -n, has the complement of bits set compared with n, but I agree that it is a nice feature and have implemented this.

cgogolin:
I have further added a unit test that checks that FlipBits can actually be applied to various forms of quregs and arrays of qubits. I think it is good to have that, as we are calling make_tuple_of_qureg() directly.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,59 @@ def get_inverse(self):
...
+
+    def __str__(self):
+        return "FlipBits("+str(self.bits_to_flip)+")"
+
+    def __or__(self, qubits):
+        for qureg in self.make_tuple_of_qureg(qubits):
+            if isinstance(self.bits_to_flip, int):
+                bits_to_flip = [((self.bits_to_flip if self.bits_to_flip >= 0 else self.bits_to_flip-1) >> i) & 1 for i in range(len(qureg))]
+            else:
+                bits_to_flip = self.bits_to_flip
```
thomashaener: This case distinction will no longer be necessary when storage is fixed to int. Also, there is no need to compute the two's complement manually: The `if flip:` below can be replaced by `if (self.bits_to_flip >> i) & 1:` without first generating the list.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,59 @@ def get_inverse(self):
...
+        elif isinstance(bits_to_flip, str):
+            self.bits_to_flip = list([ c != "0" for c in bits_to_flip])
+        else:
+            self.bits_to_flip = list(bits_to_flip)
+
+    def __str__(self):
+        return "FlipBits("+str(self.bits_to_flip)+")"
+
+    def __or__(self, qubits):
+        for qureg in self.make_tuple_of_qureg(qubits):
```
thomashaener: Is there a specific reason why FlipBits acts on each quantum register separately?

I would have either allowed a single quantum register only (with assertion), or concatenate the registers if several are provided. I think the current implementation may lead to confusion: `FlipBits(2) | (qubit1, qubit2)` vs `FlipBits(2) | (qubit1 + qubit2)`.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,59 @@ def get_inverse(self):
...
+            else:
+                bits_to_flip = self.bits_to_flip
+
+            for i, flip in enumerate(bits_to_flip):
+                if flip:
+                    XGate() | qureg[i]
+
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return self.bits_to_flip == other.bits_to_flip
```
thomashaener: With fixed storage type, we also don't have strange behavior of `__eq__` just because we initialized it with an int as opposed to a list of bits.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,59 @@ def get_inverse(self):
...
+                                    bit corresponding to qureg[0]. If bits_to_flip is negative,
+                                    exactly all bits which would not be flipped for the input
+                                    -bits_to_flip are flipped.
+        """
+        if isinstance(bits_to_flip, int):
+            self.bits_to_flip = bits_to_flip
+        elif isinstance(bits_to_flip, str):
+            self.bits_to_flip = list([ c != "0" for c in bits_to_flip])
+        else:
+            self.bits_to_flip = list(bits_to_flip)
```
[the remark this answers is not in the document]
cgogolin: Brilliant. Implemented in 6be145af8134042afbc7af039ff9cd3d3621dc38

projectq/ops/_gates.py
cgogolin: Yes, after converting to int this is the more natural and more efficient way to do this. Implemented in 6f32a1f78da2eb6740b9e5b552787cc2e05a39a2.

projectq/ops/_gates.py
cgogolin: I just mirrored the behavior of `MeasureGate`. I still would like to support all the qureg/qubit formats I test in `_simulator_test.py` and calling `make_tuple_of_qureg()` seems to be the easiest way of doing this (please correct me if I am wrong here), also this means that changes to that function will be automatically supporting.
The behavior for `FlipBits(2) | (qubit1, qubit2)` was indeed confusing and as this notation is (as far as I understand) meant to be used only for controlled gates, I now raise a ValueError if the input is a tuple of len > 1. If I try to execute `FlipBits(2) | (qubit1 + qubit2)`, I get `unsupported operand type(s) for +: 'Qubit' and 'Qubit'`, so I am not exactly sure what you had in mind here.
The tests should make it pretty clear which notations for the right hand side of `__or__()` are now supported. Let me know if you want anything further changed.

thomashaener: Great, thanks!

I was assuming that qubit1 and qubit2 were allocated via `eng.allocate_qubit()`, which returns a quantum register of length 1 and so + would concatenate the two quantum registers, resulting a register of length 2.
The notation `gate | (qureg1, qureg2)` is by no means reserved for controlled gates only. For example, addition of two registers would be `Add | (qureg1, qureg2)`.

projectq/ops/_gates.py
```diff
@@ -338,3 +338,57 @@ def get_inverse(self):
...
+
+    def __str__(self):
+        return "FlipBits("+str(self.bits_to_flip)+")"
+
+    def __or__(self, qubits):
+        if isinstance(qubits, tuple) and len(qubits) > 1:
+            raise ValueError(self.__str__()+' can only be applied to qubits, quregs, arrays of qubits, and tuples with one individual qubit')
+        for qureg in self.make_tuple_of_qureg(qubits):
+            for i, qubit in enumerate(qureg):
+                if ((self.bits_to_flip if self.bits_to_flip >= 0 else self.bits_to_flip-1) >> i) & 1:
```
thomashaener: This should just be `if (self.bits_to_flip >> i) & 1:`

cgogolin: I don't think so. The bit patterns of n and -n are not complementary, those of n and -n-1 are.

projectq/ops/_gates.py
cgogolin: Ok, I understand. Then do you prefer that I support `FlipBits(x) | (qureg1, qureg2)` and treat it in the same way as `FlipBits(x) | qureg1 + qureg2` or do you prefer the current implementation that raises a `ValueError`?

thomashaener: Either is fine imo

projectq/ops/_gates.py
thomashaener: Why do you want them to be complementary? I'd expect `FlipBits` to flip the 1-bits in the two's complement representation of a negative integer, i.e., -1 flips all bits.

projectq/ops/_gates.py
damiansteiger: I like the current restriction to a single argument and otherwise a ValueError but would be fine with either.

projectq/ops/_gates.py
cgogolin: I guess both variants "make sense", but I agree that having a gate that flips all bits is a very good argument in favor of the behavior your suggest. I made the corresponding change in  30103c7

projectq/ops/_gates.py
```diff
@@ -338,3 +338,58 @@ def get_inverse(self):
 
 #: Shortcut (instance of) :class:`projectq.ops.BarrierGate`
 Barrier = BarrierGate()
+
+
+class FlipBits(SelfInverseGate):
+    """ Gate for transforming qubits in state |0> to any computational basis state """
```
thomashaener: I think this doc string is no longer up-to-date

projectq/ops/_gates.py
```diff
@@ -338,3 +338,58 @@ def get_inverse(self):
...
+        """
+        Initialize FlipBits gate.
+
+        Example:
+            .. code-block:: python
+
+                qureg = eng.allocate_qureg(2)
+                FlipBits([0, 1]) | qureg
+
+        Note:
```
thomashaener: I think this note can be removed

projectq/ops/_gates.py
```diff
@@ -338,3 +338,58 @@ def get_inverse(self):
...
+            self.bits_to_flip = 0
+            for c in reversed(list(bits_to_flip)):
+                bit = 0b1 if c == '1' or c == 1 or c == True else 0b0
+                self.bits_to_flip = (self.bits_to_flip << 1) | bit
+
+    def __str__(self):
+        return "FlipBits("+str(self.bits_to_flip)+")"
+
+    def __or__(self, qubits):
+        if isinstance(qubits, tuple) and len(qubits) > 1:
```
thomashaener: A simple solution would be
```
quregs_tuple = self.make_tuple_of_qureg(qubits)
if len(quregs_tuple) > 1:
...
```

cgogolin: Yes, marginally simpler. Good suggestion.

projectq/ops/_gates.py
cgogolin: Yes. Done.

cgogolin:
If I move the execution tests from `_simulator_test.py` to `_gate_test.py` (that what you are asking for, right?) I also need to copy over the `sim` fixture and `get_available_simulators()`. Do you really want that or is there a way that I don't see?
I have tried to pylint my part of `_gate.py`. It now only complains about `XGate() | qubit` being assigned to nothing, but you don't seem to manually disable this in the other source files.

thomashaener:
Thanks!

Yes, this is what I meant. You can just do `eng = MainEngine()`, which will have a simulator as the default backend (C++ if available, else Python).

Assignment to nothing: That's fine. I meant running the pep8-tool. After your update, the only complaints that are left are lines being too long, backslashes that can be removed, and too many/not enough blank lines.

cgogolin:
Ah! Tests moved...

I have taken the liberty to also pep8 `_qubit_operator.py` and `_state_prep.py`.

From my side this is now done, I think. Puh, that was a lot of back and forth for a PR that I thought would be trivial. Thanks for your help and patience! It would be great if things like pep8 could be run as automatic tests on Travis....

projectq/ops/_state_prep.py
```diff
@@ -12,7 +12,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-from ._basics import BasicGate
+from ._basics import BasicGate, SelfInverseGate
```
thomashaener: Where does this `SelfInverseGate` come from?

thomashaener:
Great, thanks a lot! Yes, I agree regarding #27

projectq/ops/_state_prep.py
thomashaener: I removed it, hope that's fine.

Issue #269, opened 2018-08-31
Do not print "Waiting for results. [Job ID: ...]" when `verbose=False` on the IBMBackend


cgogolin:
When using the IBMBackend the message `Waiting for results. [Job ID: ...]` is printed for each computation, even when `verbose=False`. This clutters the output of programs using ProjectQ.

thomashaener:
Yes, but then we should also do at least one of the following
- add different levels of `verbose`
- expose the timeout
- catch Ctrl+C

So people who have been using Ctrl+C if a job is taking too long can then, e.g., adjust the timeout to a few seconds and still get the job id.

PR #294, opened 2018-12-31
Fix 502 error

cgogolin:
Fixes the 502 Server Error described in #269.


cgogolin:
After you merged 79f9af7e22e9102b1ecd5d73b6b2784892b506d2 the test started to fail, but only for python 3.6 and the error is in `_metagates.py`, which I didn't touch, so I suppose it is not my fault?

thomashaener:
No, it's not your fault; once #300 is merged, the tests should work again.

Issue #291, opened 2018-12-20
Seemingly random 502 Server Errors keep killing my quantum computations and a suggested workaround


cgogolin:
From time to time I see the following error message when using the IBM backend, and would like to know whether you are also experiencing this and maybe have an idea what is the root cause:

While running a circuit that normally executes just fine, I get the following exception log
```
../../.local/lib/python3.5/site-packages/projectq/cengines/_main.py:304: in flush
    self.receive([Command(self, FlushGate(), ([WeakQubitRef(self, -1)],))])
../../.local/lib/python3.5/site-packages/projectq/cengines/_main.py:266: in receive
    self.send(command_list)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <projectq.cengines._main.MainEngine object at 0x7fbfb2ee2ac8>, command_list = [<projectq.ops._command.Command object at 0x7fbfb2ee2630>]

    def send(self, command_list):
        """
            Forward the list of commands to the next engine in the pipeline.
    
            It also shortens exception stack traces if self.verbose is False.
            """
        try:
            self.next_engine.receive(command_list)
        except:
            if self.verbose:
                raise
            else:
                exc_type, exc_value, exc_traceback = sys.exc_info()
                # try:
                last_line = traceback.format_exc().splitlines()
                compact_exception = exc_type(str(exc_value) +
                                             '\n raised in:\n' +
                                             repr(last_line[-3]) +
                                             "\n" + repr(last_line[-2]))
                compact_exception.__cause__ = None
>               raise compact_exception  # use verbose=True for more info
E               Exception: Failed to run the circuit. Aborting.
E                raised in:
E               '  File "/home/cgogolin/.local/lib/python3.5/site-packages/projectq/backends/_ibm/_ibm.py", line 295, in _run'
[...]

PR #295, opened 2018-12-31
Logging sigint handling expose num_runs, and interval

### LINKED ISSUE #269: Do not print "Waiting for results. [Job ID: ...]" when `verbose=False` on the IBMBackend

cgogolin:
This PR closes #269 and #291 and in particular implements the following:

* `num_retries` and `interval` can now be set when initiating an `IBMBackend` and this is also explained in the docstring (this was suggested by @thomashaener in #291). The values are then passed on via `send()` and `retrieve()` to `_get_result()`. As a side effect, this makes the signatures of `send()` and `retrieve()` more similar.

* `_get_result()` now prints the "Waiting for results. [Job ID: {}]" message only if `verbose=True` and thus now handles this `print` statement like the other print statements in `_ibm_http_client.py`. This means the ProjectQ no longer clutters the output of programs using its `IBMBackend`. 

* To still allow users to interrupt the execution and retrieve a result later via the job ID, `SIGINT` is now interrupted while the IBM API is queried for results and the job ID is printed before the program exits, as was requested by @thomashaener in #269 and #291. Care has been taken to restore a potentially previously set signal handler in case one was set before exiting `_get_result()`.

Please be careful when merging this PR and #294. They both change the same portion of code and manual merging might be necessary, but as they address different issues, I thought it would be better to submit two PRs.


projectq/backends/_ibm/_ibm.py
```diff
@@ -62,6 +63,11 @@ def __init__(self, use_hardware=False, num_runs=1024, verbose=False,
             password (string): IBM Quantum Experience password
             device (string): Device to use ('ibmqx4', or 'ibmqx5')
                 if use_hardware is set to True. Default is ibmqx4.
+            num_retries (int): Number of times to retry to obtain
+                results from the IBM API. (default is 3000)
+            interval (int): Number of seconds between successive
+                attempts to obtain results from the IBM API.
+                (default is 1)
```
thomashaener: I wouldn't make this an `int`. We might even only expose `num_retries` and drop `interval` here; what do you think?

cgogolin: True, the (int) is not necessary. Fixed in 53d2483. But why not expose `interval`? Gives your users more freedom and I don't see any downsides. In fact, it makes it more obvious that the API works by periodically polling the IBM servers for a result, which is a nice side effect.

thomashaener: Because I don't see why I'd ever want to change the interval; do you?
Regarding your reason for adding it: One can also just add a comment in the docs about the "periodically polling the IBM servers" to make it obvious.

cgogolin: What makes you so sure that, say 1s, is always optimal for any application using ProjectQ? Maybe someone needs a result as soon as it is available and cannot wait for at least 1s, and maybe someone else knows that their computation will take a while and they don't want to spam IBM with requests to their API that they anyway know will yield no result.

thomashaener: I didn't claim it was optimal; I was just asking if you saw a use case.

If I cannot afford to wait for 1s, then I wouldn't use the service in the first place since the queue tends to be quite long. If I want to retrieve the results at a later point (e.g., because I know that the queue is too long), I'd just use `retrieve_execution`.

Since you find it useful to have access to `interval` (beyond it having no downsides), we should add it.

cgogolin: My perspective is that of someone developing a library on top of ProjectQ (https://pennylane-pq.readthedocs.io/en/latest/installation.html) and and in my code I anyway have to block until I get results from the queue because the following code immediately depends on the outcome. Using retreive_execution is thus not a good option, but having an option to have the user set `interval` would be nice :-)

PR #317, opened 2019-03-14
Added general-purpose VQE class/function

gavarela:
Think I should mention this is based on grove's pyQuil implementation of VQE:

https://github.com/rigetti/grove/blob/master/grove/pyvqe/vqe.py

Are there any licensing/copyrights issues?

### DIFF
--- a/examples/VQE.py
+++ b/examples/VQE.py
@@ -1,0 +1,261 @@
+## ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#  Variational Quantum Eigensolver
+## ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+import projectq as pq
+from projectq import MainEngine
+from projectq.ops import QubitOperator, Measure, All, Rx, Ry
+
+import numpy as np, cmath
+from scipy.optimize import minimize
+
+
+class VQE(object):
+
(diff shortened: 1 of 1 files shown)


mstechly:
@gavarela Grove is distributed under Apache 2.0, same as ProjectQ, so there shouldn't be any problems with that. Though, as I understand it, it should be mentioned somewhere in the code that it's based on Grove's source code.

fernandodelaiglesia:
Dear all,

I was thinking in adding a general purpose QAOA as well, and looking at the VQE proposal I have some questions I would like to discuss with you all.

- where should the functions be located? it seems that "examples" is not the more apropriate location. Somewhere as libs/ looks better, what do you think? any other place?
- I see that there is a function "expectation" in the VQE porposal that can be leveraged by the QAOA or any other custom algorithm. IMHO it is worth to make a specific function/class out of it to be used in any other place (without the need to import the full VQE). Of course if you are using the simulator you can leverage the eng.backend.get_expectation_value function, but for a real device we could leverage a function like this.
- following the documentation style, IMHO It would be great if you can add some "how to use" to the code to be included in the documentation
- I realized that you already have a QAOA in your repo that could be added to the general purpose in addition to VQE

Please let me know what do you think

Thank you very much in advance
Be

PR #324, opened 2019-05-06
Implementation of a general graph mapper for ProjectQ

Takishima:
This PR lays the ground work for an arbitrary graph mapper in ProjectQ.

Its use is similar to other ProjectQ mappers and has one required argument: a connected graph (`nx.Graph`). Here is an example:

    import networkx as nx
    from projectq.cengines import GraphMapper

    # Example of a 3x3 grid graph
    mygraph = nx.Graph()
    # Add horizontal edges
    mygraph.add_edges_from((0, 1), (1, 2),  (3, 4), (4, 5),  (6, 7), (7, 8))
    # Add vertical edges
    mygraph.add_edges_from((0, 3), (3, 6),   (1, 4), (4, 7),  (2, 5), (5, 8))

    mapper = GraphMapper(graph=mygraph)

    # use as any other mapper

Some implementation details:
- The `GraphMapper` only considers the shortest path between any two qubits when calculating a routing to update the current mapping.
- It tries to execute as many two-qubits gates simultaneously as possible, even in cases where the routing has intersecting paths for different qubits.
- It supports caching of paths to speed up calculations.

The current implementation  **does not**:
- take the directionality of gates into account (we should probably discuss if this is something the mapper has to do, or whether the backend should be the one responsible for this)
- use a weighted graph which would allow qubit routing to rely on hardware criteria (such as gate fidelity for e.g.)

### DIFF
--- a/projectq/cengines/__init__.py
+++ b/projectq/cengines/__init__.py
@@ -32,3 +32,4 @@ from ._replacer import (AutoReplacer,
 from ._tagremover import TagRemover
 from ._testengine import CompareEngine, DummyEngine
 from ._twodmapper import GridMapper
+from ._graphmapper import GraphMapper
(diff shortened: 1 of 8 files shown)


Takishima:
Superseded by #340

PR #326, opened 2019-06-03
Added an example of quantum adder

JerryZYGo:
Hi I am interested in this project, so I want to make some contribution by adding an illustrated tutorial explaining how to construct a quantum adder. I hope this can lower the learning curve for other beginners.


JerryZYGo:
> Thank you Jerry for the interest and the proposal.
> I have some concept questions, Is this adder taking advantage of the quantum nature of the qubits? If I try the notebook (using the sample test) by setting qureg_A[0] in a superposition state, which output I will get? Is the adder a binary one? if so, which is the point to have an example that is not using the quantum nature in ProjectQ?
> In addition and in order to make the notebook usable in some jupytier environment I would add at the beginning a code line with
> !pip install projectq
> Thank you

Thank you for paying attention, I am always glad to have someone to discuss with the topic of quantum computation.

As for your question,  is this adder taking advantage of the quantum nature of the qubits? I think the answer is yes (though not explicitly). This quantum adder is based on [Vedral et al.](https://arxiv.org/abs/quant-ph/9511018v1), the first implementation of quantum addition as far as I can track back in history. It is one basic building block of quantum modular exponentiation, which is the most demanding part of Shor's algorithm. So if we use it a part of Shor's algorithm, the input state will be |1>, which is a superposition of eigenstates of Multiply_Modulo_N operator. So yes, in that sense the quantum nature of parallelism is utilized.

By the way, in the build-in `add_constant` function of projectQ,  is implemented using another approach, see [Draper](http://arxiv.org/abs/quant-ph/0008033v1). However I think this one is more straightforward and easier to understand.

fernandodelaiglesia:
Sorry for the delay in the answer.

Thank you for the references and for the explanantion. Now I think it is clear the intention. I would suggest to add the information of the reference in the notebook. In addition, do you plan to add all the operations that are described in the paper? I think it would complete the example.

Regarding the comment on adding to the notebook a first code line 

`!pip install projectq`

what do you think?

Finally, if I'm not wrong, in order to be able merge, some maganement should be in place (contributor agreemnt). @Takishima, can you please help on this?

Takishima:
Everybody that contributes to ProjectQ is required to sign a Copyright Licence Agreement (CLA). The easiest way of getting it is to just send an e-mail to [info@projectq.ch](mailto:info@projectq.ch) requesting the CLA and you should then get it, read it, sign it and return it. Everything is done electronically.

Takishima:
@JerryZYGo Did you manage to get the CLA signed? If not please do so. Until this is done I will not be reviewing this PR.

PR #330, opened 2019-07-31
3 additional 2-qubit gates

dwierichs:
For convenience, these two-qubit rotation gates might be added to the gate set:
Two-qubit X rotation  exp(-i/2 \theta XX)
Two-qubit Y rotation  exp(-i/2 \theta YY)
Two-qubit Z rotation  exp(-i/2 \theta ZZ)

### DIFF
--- a/projectq/ops/_gates.py
+++ b/projectq/ops/_gates.py
@@ -16,28 +16,19 @@
 Contains definitions of standard gates such as
 * Hadamard (H)
 * Pauli-X (X / NOT)
-* Pauli-Y (Y)
+* GroundStateGate (|0><0|)
 * Pauli-Z (Z)
-* S and its inverse (S / Sdagger)
 * T and its inverse (T / Tdagger)
 * Swap gate (Swap)
-* SqrtSwap gate (SqrtSwap)
-* Entangle (Entangle)
 * Phase gate (Ph)
-* Rotation-X (Rx)
-* Rotation-Y (Ry)
 * Rotation-Z (Rz)
+* Rotation-Z on two qubits (Rzz)
 * Phase-shift (R)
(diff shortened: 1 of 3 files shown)


Takishima:
Wouldn't you be able to achieve the same result by using `All(Rx(theta)) | qureg` ?

dwierichs:
The newly implemented gates exponentiate the tensor product, being equivalent to
cos(theta/2) Id - i sin(theta/2) X_0 X_1 
whereas
exp(-i/2 theta X_0) exp(-i/2 theta X_1) = [cos^2(theta/2)-sin^2(theta/2)] Id - i cos(theta/2)sin(theta/2) [X_0 + X_1]
where the index of the Pauli X is the qubit it acts on.
is the tensor product of the exponents. I hope I did not overlook something here but then the two are indeed different.

projectq/ops/_gates.py
```diff
@@ -205,6 +217,9 @@ def matrix(self):
         return np.matrix([[cmath.exp(1j * self.angle), 0],
                           [0, cmath.exp(1j * self.angle)]])
 
+    def __str__(self):
```
Takishima: Why is this change needed? I would rather keep the old way of printing the phase gate as `"Ph(angle)"`.

dwierichs: I am sorry, I overlooked this while undoing changes I had made to the printing method.

Takishima: No worries!

Takishima:
It seems you have yet to sign the CLA agreement for contributor to ProjectQ. Could you please send an email to info@projectq.ch requesting the CLA and then sign it ?

PR #346, opened 2019-12-10
Trapped ion decomposer setup and rotation gates improvement

dbretaud:
New setup:
-> new decomposition rules for Hadamard to Rx and Ry
-> new decomposition rules for Rz to Rx and Ry
-> new decomposition rules for CNOT to Rxx, Rx and Ry
-> update on the restrictedgateset setup to allow the call of an external decomposition chooser function
-> new trapped_ion_decomposer setup calling restrictedgateset with Rx,Ry,Rxx and a new decomposition chooser function that chooses decompositions to maximize Ry gate cancellation
-> test functions for decomposition gates ignore global phases when comparing states
Improvement on Rotation gates:
-> is_identity function for every basic gates returning True if gate doesn't do anything. Ex: rotation gates with an angle being 0 or 4pi
-> local optimizer improvement to remove gates equivalent to the Identity gate
-> cosmetic improvements on the angle display for rotation gates for both python string and latex via the to_string() function


projectq/setups/decompositions/cnot2rxx_test.py
```diff
@@ -0,0 +1,139 @@
...
+# engines, respectively. 
+# The decomposition in this case only produces the 
+# same state as CNOT up to a global phase. 
+# test_vector and correct_vector represent the final 
+# wave states of correct_qb and test_qb. 
+# The dot product of correct_vector and test_vector
+# should have absolute value 1, if the two vectors are the
+# same up to a global phase.
+
+def test_decomposition():
```
Takishima: This test only tests the *_M decomposition rule. Please modify it so that it also covers the *_P decomposition rule.

To avoid duplicating code, you could parametrise this test function with a single parameter which would take one of 2 values stored in `cnot2rxx.all_defined_decomposition_rules`

Something like:
```python
@pytest.mark.parametrize("rule", cnot2rxx.all_defined_decomposition_rules)
def test_decomposition(rule):
    ...
    rule_set = DecompositionRuleSet(rules=[rule])
    ...
```

projectq/setups/decompositions/rz2rx_test.py
```diff
@@ -0,0 +1,115 @@
...
+# The decomposition only needs to produce the 
+# same state in a qubit up to a global phase. 
+# test_vector and correct_vector represent the final 
+# wave states of correct_qb and test_qb. 
+# The dot product of correct_vector and test_vector
+# should have absolute value 1, if the two vectors are the
+# same up to a global phase.
+
+@pytest.mark.parametrize("angle", [0, math.pi, 2*math.pi, 4*math.pi, 0.5])
+def test_decomposition(angle):
```
Takishima: This test only tests the *_N decomposition rule. Please modify it so that it also covers the *_M decomposition rule.

Something like:
```python
@pytest.mark.parametrize("rule", h2rx.all_defined_decomposition_rules)
def test_decomposition(angle, rule):
    ...
    rule_set = DecompositionRuleSet(rules=[rule])
    ...
```

projectq/setups/decompositions/h2rx_test.py
```diff
@@ -0,0 +1,119 @@
...
+# engines, respectively. 
+# The decomposition in this case only produces the 
+# same state as H up to a global phase. 
+# test_vector and correct_vector represent the final 
+# wave states of correct_qb and test_qb. 
+# The dot product of correct_vector and test_vector
+# should have absolute value 1, if the two vectors are the
+# same up to a global phase.
+
+def test_decomposition():
```
Takishima: This test only tests the *_M decomposition rule. Please modify it so that it also covers the *_P decomposition rule.

Something like:
```python
@pytest.mark.parametrize("rule", cnot2rxx.all_defined_decomposition_rules)
def test_decomposition(rule):
    ...
    rule_set = DecompositionRuleSet(rules=[rule])
    ...
```

projectq/setups/decompositions/rz2rx.py
```diff
@@ -0,0 +1,68 @@
...
+    angle = cmd.gate.angle
+
+    with Control(eng, cmd.control_qubits):
+        with Compute(eng):
+            Ry(math.pi/2.) | qubit
+        Rx(angle) | qubit
+        Uncompute(eng)
+
+def _recognize_RzNoCtrl(cmd):
+    """ For efficiency reasons only if no control qubits."""
```
Takishima: What would be the "inefficient" case?

projectq/setups/trapped_ion_decomposer.py
```diff
@@ -0,0 +1,155 @@
...
+    for el in decomposition_list:
+        try:
+            decomposition=el.decompose.__name__.split('_')
+            name=decomposition[2]
+            decomp_rule[decomposition[3]]=el
+            # 'M' stands for minus, 'P' stands for plus
+            # 'N' stands for neutral
+            # e.g. decomp_rule['M'] will give you the 
+            # decomposition_rule that ends with a Ry(-pi/2)
+        except:
```
Takishima: Avoid naked `except` statements. In this case, you probably want to catch `IndexError` exceptions.

projectq/setups/trapped_ion_decomposer.py
```diff
@@ -0,0 +1,155 @@
...
+            pass
+    if name=='cnot2rxx': 
+        ctrl = cmd.control_qubits
+        idx=str(ctrl) # index of the qubit
```
Takishima: Rather than converting the list of qubits to a string, I would use the qubit identifiers:

```python
idx = [qb.id for qb in cmd.control_qubits]
```

projectq/setups/trapped_ion_decomposer.py
```diff
@@ -0,0 +1,155 @@
...
+                return decomp_rule['P']
+        else:
+            # In this case where we have no previous Ry sign 
+            # we choose decomp_rule['M']
+            prev_Ry_sign[idx]=-1
+            return decomp_rule['M']
+    elif name=='h2rx':
+        # Block operates similar to 'cnot2rxx' case
+        qubit = cmd.qubits[0]
+        idx=str(qubit)
```
Takishima: Rather than converting the list of qubits to a string, I would use the qubit identifiers:

```python
idx = [qb.id for qb in cmd.qubits[0]]
```

projectq/setups/trapped_ion_decomposer.py
```diff
@@ -0,0 +1,155 @@
...
+        elif prev_Ry_sign[idx]==0:
+            prev_Ry_sign[idx]=+1
+            return decomp_rule['M']
+        else:
+            prev_Ry_sign[idx]=00
+            return decomp_rule['N']
+    elif name=='rz2rx':
+        # Block operates similar to 'cnot2rxx' case
+        qubit = cmd.qubits[0]
+        idx=str(qubit)
```
Takishima: Rather than converting the list of qubits to a string, I would use the qubit identifiers:

```python
idx = [qb.id for qb in cmd.qubits[0]]
```

projectq/setups/trapped_ion_decomposer.py
```diff
@@ -0,0 +1,155 @@
...
+        -Classical instructions gates such as e.g. Flush and Measure are
+        automatically allowed.
+        -The restricted gate set engine does not work with Rxx gates, as 
+        ProjectQ will by default bounce back and forth between Cz gates 
+        and Cx gates. An appropriate decomposition chooser needs to be used!
+
+    Returns:
+        A list of suitable compiler engines.
+    """
+    return restrictedgateset.get_engine_list(one_qubit_gates=(Rx,Ry),two_qubit_gates=(Rxx,),compiler_chooser=chooser_Ry_reducer)
```
Takishima: Please wrap line to 80 characters.

projectq/ops/_basics.py
```diff
@@ -317,19 +322,19 @@ class BasicRotationGate(BasicGate):
     A rotation gate has a continuous parameter (the angle), labeled 'angle' /
     self.angle. Its inverse is the same gate with the negated argument.
     Rotation gates of the same class can be merged by adding the angles.
-    The continuous parameter is modulo 4 * pi, self.angle is in the interval
-    [0, 4 * pi).
+    The continuous parameter is modulo 2 * pi, self.angle is in the interval
+    [0, 2 * pi).
```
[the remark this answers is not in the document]
dbretaud: Thank you for your feedback. We spent some time thinking about your comment regarding global phase and have concluded that the global phase should not affect the measurements of our final states. The global phase acts simultaneously on all the amplitudes of our qubit states without however changing the final measurement probability for any outcome, and can therefore be discarded. Our conclusion is further supported by these papers mentioning gate decomposition [1][2].

[1]https://iopscience.iop.org/article/10.1088/1367-2630/aa5e47 
[2]https://journals.aps.org/pra/pdf/10.1103/PhysRevA.79.012312

projectq/setups/decompositions/rz2rx.py
dbretaud: That's actually a good question, we just kept the same comment here as for the rx2rz.py and ry2rz.py files.

projectq/ops/_basics.py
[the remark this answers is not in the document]
Takishima: It is true that for single-qubit gates, the global phase can be safely ignored, but there are some cases where this global phase cannot be ignored: e.g. controlled single qubit gates where the global phase turns into a local phase that should not be ignored.

projectq/cengines/_optimize.py
```diff
@@ -130,6 +130,23 @@ def _optimize(self, idx, lim=None):
...
+                                for qb in sublist]
+                    gid = self._get_gate_indices(idx, i, qubitids)
+                    for j in range(len(qubitids)):
+                        new_list = (self._l[qubitids[j]][0:gid[j]] +
+                                    self._l[qubitids[j]][gid[j] +1:])
+                        self._l[qubitids[j]] = new_list
+                    i = 0
+                    limit -= 1
+                    continue
+            except:
```
Takishima: Avoid naked `except` block.

Takishima:
I had to fix a few issues here and there to get all the tests to pass and make sure the test coverage was proper, but I will now merge those changes.

Many thanks for taking the time to implement new features into ProjectQ and for signing the CLA.

PR #353, opened 2020-02-04
Aqt Backend

dbretaud:
New Backend to connect to the Alpine Quantum Technologies (AQT) devices.

-> A documentation was released mid-February:
https://www.aqt.eu/aqt-json-tutorial/#json-syntax
-> Access keys can be requested by creating an account at: https://gateway-portal.aqt.eu

->Only tested on the online simulator but the code should be the same for the hardware
->This backend is a trapped ion quantum computer so only accepts Rx,Ry and MS gates

### DIFF
--- a/projectq/backends/__init__.py
+++ b/projectq/backends/__init__.py
@@ -24,9 +24,11 @@ This includes:
 * a resource counter (counts gates and keeps track of the maximal width of the
   circuit)
 * an interface to the IBM Quantum Experience chip (and simulator).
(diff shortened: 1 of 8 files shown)


dbretaud:
Looks good to me. There is no real need for mapper as the trapped ion machines have usually full connectivity but I guess it could be useful for later updates

Takishima:
Thanks for fixing the issues I missed and as usual for your work.

PR #376, opened 2020-08-12
Update requirements.txt

SiddTim:
Added seaborn to requirments.txt


Takishima:
I do not see the point of adding `seaborn` to the requirements for ProjectQ as we are not using it internally at any point.

Issue #378, opened 2020-09-14
REST API update


Sov-trotter:
I have been referring to the API client in this repo, to build the a similar thing in The Julia Language. According to https://github.com/ProjectQ-Framework/ProjectQ/blob/develop/projectq/backends/_ibm/_ibm_http_client.py#L165-L169, and the docs too, there have to be three fields in the response, viz. `downloadQObjectUrlEndpoint`, `uploadQobjectUrlEndpoint` and `uploadUrl`, but I have been getting only the `uploadUrl` in response. 
I happened to discuss it with one of the lead devs of IBMQ and he confirmed that there had indeed been an update, where there's isn't any need of the extra step and directly `PUT` the QObject to the `uploadUrl`.

I didn't try the code here in Python, but if someone can confirm then it'd be nice. :)

dbretaud:
Indeed, I can see now there has been some (additional) modifications to the API since ProjectQ was updated. I will take a look into that

Sov-trotter:
[Din't know where else to ask it] 
Hey @dbretaud I am trying to get the API worked out in Julia and it's abit hard to understand the format of the `Dict` (Qobj) that is being sent. Can you share an example?

PR #379, opened 2020-09-15
update on the ibmq client

dbretaud:
Update the ibmq API as noticed in #378


Takishima (approved):
As always thanks for the contribution.

PR #389, opened 2021-02-23
Update .travis.yml configuration & add .coveragerc

Takishima:
- Added .coveragerc to ignore test files in coverage
- Cleaned up .travis.yml
- Fixed a couple of bugs leading to test failures
- Added build on Mac OS X (Homebrew)
- Added build on Windows
- Removed Python 2.7 & 3.4 builds as this version has reached its end-of-life

### DIFF
--- a/projectq/backends/_sim/_pysim.py
+++ b/projectq/backends/_sim/_pysim.py
@@ -20,6 +20,11 @@ Please compile the c++ simulator for large-scale simulations.
 
 import random
 import numpy as _np
+import os
+
+_USE_REFCHECK = True
+if 'TRAVIS' in os.environ:
+    _USE_REFCHECK = False
 
 
 class Simulator(object):
@@ -110,7 +115,7 @@ class Simulator(object):
         """
         self._map[ID] = self._num_qubits
         self._num_qubits += 1
-        self._state.resize(1 << self._num_qubits)
+        self._state.resize(1 << self._num_qubits, refcheck=_USE_REFCHECK)
 
     def get_classical_value(self, ID, tol=1.e-10):
         """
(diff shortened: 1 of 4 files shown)


projectq/backends/_sim/_pysim.py
```diff
@@ -20,6 +20,11 @@
 
 import random
 import numpy as _np
+import os
+
+_USE_REFCHECK = True
+if 'TRAVIS' in os.environ:
+    _USE_REFCHECK = False
```
andreashehn, line 27: Why do you set refcheck `false` for travis?

Takishima: This is because of some errors that seem to happen only when we run the coverage tests with pytest. I am assuming this happens because the coverage tests injects some code somewhere to properly keep track of the code coverage status and that may interfere with some of numpy reference checking mechanisms.

```
ValueError: cannot resize an array references or is referenced
by another array in this way.  Use the resize function
```

However, I have not been able to find any helpful documentation so far to properly explain that behaviour. Plus I can't seem to reliably reproduce this behaviour on my machines at the moment.

As far as I can remember, running the tests without the coverage plugin does not lead to the same error, so I am reluctant to entirely disable the feature by passing `refcheck=False` to `resize(...)` since that might still help catch some unwanted behaviour.

Takishima: One other way to solve that issue would be using the free function `numpy.resize(...)` instead of the array method, but that would probably induce more allocation/deallocations of memory leading to some worse performance.

andreashehn: ok, makes sense then.

andreashehn (approved):

PR #400, opened 2021-06-14
Ctrl generalize

XYShe:
Add support for off-control


Takishima (approved):
Looks good to me; we anyway have been discussing this outside of GitHub for a while now. There are just a couple of CI errors that need investigating before merging.

Takishima (approved):

PR #417, opened 2021-11-10
Daisyrainsmith/commutation2

daisyrainsmith:
Adding commutation logic 
Rewrite of PR # 386

### DIFF
--- a/projectq/cengines/_optimize.py
+++ b/projectq/cengines/_optimize.py
@@ -17,30 +17,31 @@
 
 import warnings
 
-from projectq.ops import FastForwardingGate, FlushGate, NotMergeable
-
-from ._basics import BasicEngine
-
+from projectq.cengines import BasicEngine
+from projectq.ops import FlushGate, FastForwardingGate, NotMergeable, XGate
+from projectq.ops._basics import Commutability
 
 class LocalOptimizer(BasicEngine):
     """
-    Circuit optimization compiler engine.
-
-    LocalOptimizer is a compiler engine which optimizes locally (merging rotations, cancelling gates with their
-    inverse) in a local window of user- defined size.
-
-    It stores all commands in a dict of lists, where each qubit has its own gate pipeline. After adding a gate, it
-    tries to merge / cancel successive gates using the get_merged and get_inverse functions of the gate (if
-    available). For examples, see BasicRotationGate. Once a list corresponding to a qubit contains >=m gates, the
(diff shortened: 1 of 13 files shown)


Takishima:
There are a lot of failing tests that you should probably fix before I will review this again.

daisyrainsmith:
Hi Takishima, 

Could you be more specific? When I run tests on my computer, **ops** and **cengines** pass all tests. These are the only folders which I expect to be affected by the new code. There are a lot of tests that fail that I can't see any relation to my new code, such as: **libs/math/_gates_math_test** and **backends/_aqt/_aqt_test**. Could you tell me which tests fail for you that you think are related to the new code?

Thanks, 

Daisy.

Takishima:
Ok, so here is a small breakup of the CI test failures:

# Static analysis

Link: https://github.com/ProjectQ-Framework/ProjectQ/actions/runs/1445382789

Since the last major code update, ProjectQ is now completely checked using static analysis tools like:
- `clang-tidy`
- `black`
- `isort`
- `flake8`
- `pylint`

Some of the issues can be fixed very easily on your side (mainly all the issues linked to formatting discrepancies). In order to solve those, simply do the following:
- Install `pre-commit` (https://pre-commit.com/)
- Run pre-commit from within the root directory of your ProjectQ installation:
   
      pre-commit run --all-files --hook-stage manual

- Commit the changes

For the other checks (ie. mainly linters such as `flake8` and `pylint`) you will need to address the issues one by one. The error messages [here](https://github.com/ProjectQ-Framework/ProjectQ/runs/4168883343?check_suite_focus=true) should be pretty explanatory.

My suggestions for errors due to too many statements or too many branches (`pylint`), try to remove them by restructuring the code if you are close to the limit, but. if it is not feasible or not reasonable, you may add a comment to tell the tool to ignore the warning for the particular case.

# Unit tests

Link: https://github.com/ProjectQ-Framework/ProjectQ/actions/runs/1445382791

In these tests, some of the errors are indeed quite puzzling and I did not have the time to investigate them in much details over the last few days, but I'll try to dive deeper next week if I find some time.

For sure, the errors about `IndexError`s located in `_command.py` at line 222 looks very suspicious and you may ignore them for now.

Some others are probably linked to the fact that the new optimizer is enabling commutation by default and so therefore some of the assumptions made when writing tests are not satisfied anymore. This is most likely the case of the `test_chooser_Ry_reducer` test.

PR #449, opened 2022-10-20
Context Manager: with flushing(MainEngine()) as eng:

cclauss:
Simplify the creation, use, and flushing of any Engine using a Python context manager.

When exiting the `with` block, the engine is _automatically_ flushed.

Just like auto-closing of files:
```
with open("my_file.txt") as in_file:
    print(in_file.read())
 ```
Modeled after https://docs.python.org/3/library/contextlib.html#contextlib.closing
```python
from projectq import MainEngine, flushing  # import the main compiler engine
from projectq.ops import (
    H,
    Measure,
)  # import the operations we want to perform (Hadamard and measurement)

with flushing(MainEngine()) as eng:  # create a default compiler (the back-end is a simulator)
    qubit = eng.allocate_qubit()  # allocate a quantum register with 1 qubit

    H | qubit  # apply a Hadamard gate
    Measure | qubit  # measure the qubit

    # This line is no longer required... eng.flush()

print(f"Measured {int(qubit)}")  # converting a qubit to int or bool gives access to the measurement result

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [vNext] - 2022-mm-dd
+## [Unreleased]
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [vNext] - 2022-mm-dd
+
+### Added
+
+-   Python context manager `with flusing(MainEngine()) as eng:`
+
 ## [v0.8.0] - 2022-10-18
 
 ### Added
--- a/projectq/cengines/__init__.py
+++ b/projectq/cengines/__init__.py
@@ -14,6 +14,8 @@
 
 """ProjectQ module containing all compiler engines."""
 
+from contextlib import contextmanager
+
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: projectq/cengines/__init__.py [hunk 1]
```diff
@@ -33,3 +35,11 @@
 from ._tagremover import TagRemover
 from ._testengine import CompareEngine, DummyEngine
 from ._twodmapper import GridMapper
+
+
+@contextmanager
+def flushing(thing):
```
Takishima: Missing docstring for this function.

Also, I would name the argument `engine` in order to make clear to users that a compiler engine is expected.

Lastly, please add at least one simple unit test that makes sure that this does indeed calls `flush()`. You could for example create a dummy engine class that derives from `DummyEngine` and override its `flush` method to track whether it was called.

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [vNext] - 2022-mm-dd
```
Takishima: Please replace this line with this one (that's actually a mistake I made when releasing the latest version):

`## [Unreleased]`

Takishima (changes requested):
All in all, a good addition. I just have one major comment that I would like to see addressed.

Takishima:
Also, if you have not done so already, please send an email to info@projectq.ch to request the ProjectQ license agreement that you'll have to sign before we can merge this PR.

cclauss:
% `pytest`
```
...
projectq/cengines/_withflushing_test.py ..                               [ 58%]
...
```

Takishima (approved):

Takishima:
Looks good to me. Now before I can merge this, I do need to have you sign the ProjectQ CLA.

EDIT: I just saw your email in the ProjectQ mailbox and the person responsible for it will contact you shortly.

Takishima:
Thanks for the contribution!

PR #465, opened 2023-07-24
Open v0.3 API + debiasing

splch:
This new version of the API separates job status and metadata from actual results into two different endpoints.

v0.3 also includes support for error_mitigation settings like symmetrization as described in https://arxiv.org/pdf/2301.07233.pdf

This includes:

* add a new error_mitigation parameter on job submission so users can configure it
* add a new aggregation parameter on results that would allow getting results aggregated under the two different methods described in the paper

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Fixed some typos (thanks to @eltociear, @Darkdragon84)
 
+### Changed
+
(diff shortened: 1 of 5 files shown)


Takishima:
Hi,
Thank you for submitting this PR. I will try to review it this week but the CI failures will need to be addressed before this can be merged.

splch:
> Hi, Thank you for submitting this PR. I will try to review it this week but the CI failures will need to be addressed before this can be merged.

hey! i got the ionq tests updated :) thanks

